PR #253, opened 2023-12-21
Fix typos in README and add more guidance

awunderground:


cdsolari (approved):
Thanks for enhancing this, Aaron! I think this should help!

tinatinc (approved):

PR #261, opened 2024-01-05
Consolidate fixes to income and employment metrics

jwalsh28:


jwalsh28:
@awunderground The updated files in the geographic-crosswalk folder are what Amy is reviewing now but I used the output in the ACS programs so I included them here. You can abstain from reviewing these files as Amy and I will likely be making updates.

jwalsh28:
Added a text file to the raw folder in data/temp because I could not get Git to create the folder while it was emtpy. This is where the API function stores the IPUMS extracts.

jwalsh28:
Consolidated changes in pull request #261

PR #270, opened 2024-02-01
Iss215

JCarterUI:
Review for 8th grade ELA metric


cdsolari (changes requested):
I left comments within your code that should address all the following issues that I noticed, summarized below:
• SEDA_all_county.csv is empty
• SEDA_all_subgroups_county: 
	• Change the order of variables. It should lead with year- follow instructions in the ReadMe
	• Drop: county_5digit, _megre_missing_outcome
	• Variable names must be fixed. The naming convention originally prescribed follows a convention across all metrics. Please rename "achivement_change" to "rate_learning" and the other variables with the _lb, _ub and _quality suffixes should all lead with "rate_learning"
	• In the subgroup county file, the subgroup "All" should have a corresponding "subgroup_type" = "all"
	• I see that we have a lot of negative numbers, and way more than the handful of negative values we had from 2013-2017. Is that expected? - clarified with Jay that the metric is more of a difference in mean scores between 2022 and 2019, so negative means you performed worse in 2022 than in 2019 and positive is that you performed better. These are age-adjusted and standardized test scores, so having no change is that you're performing the same, and a more positive value is that you're performing better. Is that correct?
	• Note that because these are 2022 data, it means we have missing values for the state of CT. This is a known limitation. 
• SEDA22_all_subgroups_city.csv
	• Same as above, but for the drop list it should be: city_name, state_name, _merge, missing_outcome
Same issue with variable names, ordering of variables and the missing subgroup_type="all"

awunderground:
The data do not support our planned analysis. We are reverting to the earlier years of data.

PR #277, opened 2024-02-14
Update 2021 college readiness metric, add confidence intervals and de…

jwalsh28:
…lete old files

1) Update college readiness programs for place and county level with 2021 data
2) Tranistion files to QMD and add internal quality checks/improve documentation
3) Add confidence intervals


08_education/college_metrics_place.qmd
```diff
@@ -0,0 +1,1116 @@
...
+metrics_education_all %>% 
+  select(share_hs_degree:share_hs_degree_quality) %>% 
+  summary()
+
+metrics_education_sub %>% 
+  select(share_hs_degree:share_hs_degree_quality) %>% 
+  summary()
+```
+
+Look at distribution of hs_degree share by place.
```
wcurrangroome, line 942: Opportunity to put this into loop (maybe using purrr::walk)?

wcurrangroome (commented):
This looks great overall. Pretty complex data and analysis given the need to use replicate weights, size of datasets, etc., and I see virtually no substantive issues. I think the only major issue is that the `as_survey_design` function does not work with replicate weights (to the best of my understanding; I think it just ignores the replicate weights-related arguments); this should be changed to `as_survey_rep`. 

The remaining comments relate to style, organization, and opportunities to consolidate / function-alize code and are in large part my preference; I'd encourage you to review these, assess whether they make sense to you (and how much time they'll take to address/adopt), and then make corresponding changes or just ignore or save for a subsequent year's updates. High-level, some of these suggestions include:

- Is it worthwhile to organize the parallel pieces of the subgroups- and all-data analyses into combined chunks so that it's clearer that you're applying the same logic to each of these datasets?
- Is it possible to consolidate chunks where they operate on the same underlying data? For example, in the data cleaning phases, I would combine filter() and mutate() and select() and rename() statements into a single pipe-chain in a single chunk. Then, in a subsequent chunk(s), include visuals / tables / other tests that validate or illustrate the preceding operations. Relatedly--I'm very leery of writing-over an object repeatedly (e.g., as is done with acs_clean), and would suggest naming intermediate objects, e.g., acs_1, acs_2, and then only naming the final object acs_clean. Consolidating chunks into fewer (or a single) pipe-chain reduces the need for intermediate objects.

Could you review this feedback, make any changes to this script and parallel changes to the county-level script, and then I can tag back in and review both? (I've reviewed the county-level file quickly, but it's so similar to this one, I haven't replicated all these comments for that script--most is not all should apply there as well).

wcurrangroome (changes requested):
Updating status of the review to "Request changes"

wcurrangroome (changes requested):
This look ~98% there. All instances of `acs_survey_subgroups` should be renamed to `acs_survey_all` since `acs_survey_subgroups` no longer exists. I added a few other minor comments that are not a high priority. If you can update those object names and then run top-to-bottom to ensure that everything runs correctly--for both the place- and county-level files--these are good to go!

08_education/college_metrics_county.qmd
```diff
@@ -0,0 +1,1171 @@
...
+```
+
+
+## Clean Data
+
+Missing data is reported in a variety of ways by IPUMS data. This step walks through the missing values in key variables and checks that we are dealing with them appropriately. 
+
+Age: No missing values reported in variable description. Check the distribution to confirm there are no abnormalities: the distribution of age looks reasonable (no bunching or extreme outliers).
+
+```{r}
```
wcurrangroome, line 126: No error, but this is a density plot of just ages 19 and 20, since filtering by age now occurs at the top of this script.

08_education/college_metrics_county.qmd
```diff
@@ -0,0 +1,1171 @@
...
+  theme_minimal() +
+  ggtitle("Age Variable Results") +
+  ylab("Density")
+```
+
+EDUCD: N/As reported as "1" in all modern surveys. N/As were "000" in the 1980 and 1950 decennial census surveys (not used here).
+Variable codes explained [here]([https://usa.ipums.org/usa-action/variables/EDUC#codes_section) & see "Detailed codes" at bottom).
+Vacancy is reported as "0". 
+
+There is a significant share of records reported as "1" for the education variable.
```
wcurrangroome, line 139: I don't think this is true anymore given changes to survey samples and earlier subsetting to relevant ages.

08_education/college_metrics_county.qmd
```diff
@@ -0,0 +1,1171 @@
...
+  ggplot(mapping = aes(x = factor(educd), y = n)) +
+  geom_col() +
+  theme_minimal() +
+  ggtitle("Educd, All Records") +
+  scale_y_continuous(expand = expansion(mult = c(0, 0.1))) +
+  labs(x = "Educational Attainment",
+       y = NULL)
+```
+
+Confirm the missing values from the value labels of the educd variable to ensure we are eliminating the correct values. "0", "1" and "999" are N/A or missing and should be turned into NA values that R recognizes.
```
wcurrangroome, line 154: Similar to comments above, there are no missing categories to begin with.

08_education/college_metrics_county.qmd
```diff
@@ -0,0 +1,1171 @@
...
+Turn the missing/"not in universe" value from the educd variable into NAs and remove the labels for the rest of the analysis. Rename to the more meaningful "education_detailed". 
+```{r}
+acs_clean <- acs_combined %>%
+  mutate(education_detailed = zap_labels(educd), 
+         education_detailed = ifelse(education_detailed %in% c(1, 0, 999), NA_integer_,
+                          education_detailed)
+  )
+```
+
+Look at the distribution of education_detailed after adjustment for NAs. We can see that there are no longer records for "1" but we now see "NA".
```
wcurrangroome, line 171: Reference comments above; there are no "NA" values at this point because there were not missing categories to begin with.

08_education/college_metrics_county.qmd
```diff
@@ -0,0 +1,1171 @@
...
+## Crosswalk
+
+Read in the PUMA to county crosswalk file. This file is created by the program generate_puma_county_crosswalks.rmd in the geographic-crosswalks folder. 
+```{r}
+puma_county_crosswalk <- read_csv(here::here("geographic-crosswalks", "data", "crosswalk_puma_to_county.csv"))
+```
+
+## Merge Crosswalk 
+
+Create a variable in the ACS data indicating whether the survey is pre or post-2022. This will be used to join on the correct crosswalk information.
```
wcurrangroome, line 504: Just a note that, with the change in datasets to 5-year estimates, nothing has a `crosswalk_period` of "2022" at this point, right?

08_education/college_metrics_county.qmd
```diff
@@ -0,0 +1,1171 @@
...
+
+Confirm there are no vacant properties included in the data.
+```{r}
+acs_crosswalked %>%
+  filter(is.na(vacancy)) %>% 
+  group_by(pernum) %>% 
+  count()
+```
+
+## Create pre-K metric: All
```
wcurrangroome, line 590: Here and throughout college metrics scripts, CTRL-F and replace all "pre-K" with "college readiness"--there are a few instances of this.

wcurrangroome (changes requested):
This looks great, especially the new functions! I left a few comments about text changes so that visuals / data cleaning process descriptions align with the underlying data after the change to only using 5-year estimate data. There should be no need to re-generate the output files. All comments apply to both county- and place-level files. After that--good to merge!

jwalsh28:
@wcurrangroome I think this is just about there. Made chnages based on the last round of comments with the most signficant being adding tests on the count of rows in the expanded files so that we know all UMF places and counties are included.

wcurrangroome (approved):

PR #284, opened 2024-02-22
Iss224 -- home values

malcalakovalski:
#### Ratio of the share of a community’s housing wealth held by a racial or ethnic group to the share of households of the same group

- Add 2022
- Rename variables
- Add age subgroup

**Note:** I get 3,146 counties in 2022 for this metric instead of 3,143. Extra reviewer attention on the crosswalk section would be extremely helpful to debug this.


01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+
+## Housekeeping {#sec-housekeeping}
+
+```{r}
+#| label: setup
+
+options(scipen = 999)
+librarian::shelf(tidyverse, ipumsr, here, glue, labelled, skimr, urbnthemes, tidylog, assertr)
+set_urbn_defaults(style = "print")
+```
```
kmartinchek, line 47: For folks who don't have urbnthemes loaded, line 46 will not work. Recommend removing or including instructions for download, as librarian::shelf will not work for this package.

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+
+### Read Data {#sec-read-data}
+
+We extract the 1 year ACS files from the IPUMS API using `ipumsr`. You will need to have an API key in order to do so. If you don't have one, you can register to the [IPUMS USA](https://uma.pop.umn.edu/usa/user/new?return_url=https%3A%2F%2Fusa.ipums.org%2Fusa-action%2Fmenu) data collection and create an API key.
+
```
kmartinchek, line 52: There are a lot more steps to this (e.g., apply to use the data to have permission to use your API key, etc)-- how much detail do we think is needed here?

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+years <- 2014:2022
+# usa_ext_def <- define_extract_usa(
+#   description = "ACS 1 Year 2014-2022",
+#   samples = paste0("us", years, "a"),
+#   variables = acs_vars
+# )
+#
+# usa_ext_submitted <- submit_extract(usa_ext_def)
+# usa_ext_complete <- wait_for_extract(usa_ext_submitted)
```
kmartinchek, line 76: Line 75 and beyond is not replicable to me at this time because I get invalid variable name (API request fails with status 400) for state and county. If you look on the IPUMS site, it shows it should be statefip and countyfip. When I change this, I can successfully submit the extract.

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+
+```{r}
+#| label: load-acs
+#| cache: true
+#| dependson: "import-acs-one-year"
+#| include: false
+
+# Set microdata to whatever the file you downloaded in the previous chunk is called
+microdata <- "usa_00010.xml"
+
```
kmartinchek, line 98: I would add a super-clear note in the text above this code that line XX (currently 97) would need to be manually changed to the extract number the analyst gets from the code.

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+
+acs1yr <-
+  acs1yr_raw |>
+  mutate(
+    valueh = lbl_na_if(valueh, ~ .lbl %in% c("Missing", "N/A")),
+  )
+
+acs1yr |>
+  assertr::verify(valueh < 9999998 | is.na(valueh))
```
kmartinchek, line 132: It took a bit to understand how this handles the missing values exactly, from a code logic perspective (since the function lbl_na_if comes from ipumsr (https://rdrr.io/cran/ipumsr/man/lbl_na_if.html), just noting this.

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+```
+
+Next, we exclude individuals living in group quarters because we only want households for this metric. In particular, we exclude people living in places other than housing units (e.g., correctional facilities, nursing homes, mental hospitals, dorms, barracks, group homes, missions). For more information, see the [variable description](https://usa.ipums.org/usa-action/variables/GQ#codes_section) on IPUMS USA.
+
+```{r}
+#| label: keep-households-only
+acs1yr <-
+  acs1yr |>
+  filter(!(gq %in% c("Group quarters--Institutions", "Other group quarters")))
```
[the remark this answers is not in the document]
kmartinchek: I also wonder here if code 5 should also be included, as this falls under group quarters in the IPUMS codes: https://usa.ipums.org/usa-action/variables/GQ#codes_section

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+acs1yr <-
+  acs1yr |>
+  mutate(
+    across(c(state, county, city, puma), as.character),
+    state = str_pad(state, width = 2, pad = "0", side = "left"),
+    county = str_pad(county, width = 3, pad = "0", side = "left"),
+    city = str_pad(city, width = 5, pad = "0", side = "left"),
+    puma = str_pad(puma, width = 5, pad = "0", side = "left")
+  )
+```
```
kmartinchek, line 158: After running this code chunk, I look at acs1yr and these variables look numeric, not string and do not have leading zeros.

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+
+The weight of each household is adjusted by the area of the PUMA that falls into a given County/Place
+
+```{r}
+#| label: define-crosswalk-years
+
+acs1yr <-
+  acs1yr |>
+  mutate(crosswalk_period = if_else(year < 2022, "pre-2022", "2022"))
+```
```
kmartinchek, line 215: This may be the start of where you get counting issues later on-- the CT county change that impacts the 2022 data: 
image

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+puma_county_crosswalk <-
+  read_csv(here::here(
+    "geographic-crosswalks",
+    "data",
+    "crosswalk_puma_to_county.csv"
+  )) |>
+  rename(
+    state = statefip,
+  ) |>
+  filter(state != "72")
```
kmartinchek, line 232: I'd mention you're filtering out PR in the text introducing this code.

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+  ggplot(aes(x = share_households)) +
+  geom_histogram(color = "blue", fill = alpha("blue", 0.3), binwidth = 0.02) +
+  facet_wrap(. ~ race_ethnicity) +
+  labs(
+    title = "Share of Households by Race and Ethnicity",
+    subtitle = "County",
+    y = "Count",
+    x = "Share of Households"
+  )
+
```
kmartinchek, line 337: This visualization shows up in the "plots" window-- I'm not sure if you wanted the results embedded in the .qmd or not?

kmartinchek: The other visualizations also show up in the plot window-- some are harder to see this way than others. Maybe note that all plots will be shown there?

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+-   Data quality is marginal (`2`): N is greater than 30 but only 35% to 75% of observations are from the county/place.
+-   Data quality is poor (`3`): N is less than 30 **or** if less than 35% of observations are from the county/place
+
+```{r}
+#| label: quality-check-flag-functions
+
+construct_quality_flag <- function(x, y) {
+  case_when(
+    x < 30 ~ 3,
+    x >= 30 & y < 0.35 ~ 3,
```
kmartinchek, line 485: The text above this suggests that line 485 should be y < 0.35 only -- since it uses "or". If you mean "Data quality is poor (`3`): N is less than 30 **or** if **there are more than 30 observations but** less than 35% of observations are from the county/place" (as in the code as written is accurate)-- I would update the text above this code chunk.

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+
+### County
+
+```{r}
+#| label: add-quality-check-flags-county
+
+county_quality <-
+  households_county |>
+  distinct(year, state, county, .keep_all = TRUE) |>
+  select(year, state, county, sum_products)
```
kmartinchek, line 502: sum_products is from the crosswalk-- I might note where the crosswalk documentation can be found in the text introducing the code where you start using the crosswalk, so folks can look up what this variable is.

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+  reshape_to_wide(place)
+
+ratio_house_value_households_place_subgroup <-
+  ratio_house_value_households(
+    share_households_place_subgroup,
+    house_value_place_subgroup,
+    "subgroup"
+  ) |>
+  reshape_to_wide(place, subgroup)
+```
```
kmartinchek, line 769: For these merges, there was no information on whether you expected perfect matches (i.e., all observations to merge) for these two datasets. I would add unit tests to confirm unmatched data from x is what you expected to be unmatched (as a failsafe) and also add text above the matches to describe why the matches are working as expected.

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+ratio_house_value_households_place_subgroup |>
+  mutate(subgroup_type = "age", .before = "subgroup") |>
+  bind_rows(ratio_house_value_households_place |>
+    mutate(
+      subgroup_type = "all",
+      subgroup = "All"
+    )) |>
+  arrange(year, state, place, subgroup_type, subgroup) |>
+  export_metric(file_type = "subgroup")
+```
```
kmartinchek, line 904: With the changes I made (noted in prior comments), this file reproduces the final output.

I noted, however, several things in the final output: 
1. There are some NAs in the quality flags variables that I am unsure is expected based on the code generating such flags earlier, Are these missing values expected? If not, this should be examined again.
2. Several place final csv files including the all_city and the race_ethnicity_subgroup_city files have all NA values for state_name and place_name variables. This was not an issue in the 2022_city file.

Please let me know if you have any questions about any of my comments.

01_financial-well-being/housing-value.qmd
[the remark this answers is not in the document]
malcalakovalski: @awunderground Do you know if 5 (Additional households under 2000 definition) should be included here?

01_financial-well-being/housing-value.qmd
malcalakovalski: I think this will depend on each user's RStudio settings.
Screenshot 2024-03-05 at 1 49 47 PM

01_financial-well-being/housing-value.qmd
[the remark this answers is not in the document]
awunderground: I would ask JP Walsh or Jung Choi. They calculated the original metric.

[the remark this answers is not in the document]
malcalakovalski: @jwalsh28 do you remember the motivation for excluding 5? or should it have been included originally?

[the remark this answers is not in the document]
malcalakovalski: Also, @kmartinchek I just re-ran the code with the value labels and it worked. Maybe we have different versions of `ipumsr`? I have version 0.7.1, does updating solve this?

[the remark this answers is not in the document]
jwalsh28: I don't think there was a justification for excluding 5, we are including is in the household income metric and I think it should be included here.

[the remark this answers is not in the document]
malcalakovalski: Okay, I'll make sure to filter it out in that case. Thank you!

01_financial-well-being/housing-value.qmd
malcalakovalski: I confirmed and added a test that only new FIPS are present in 2022. However, only 6 out of the 9 were present. So I think you're right that this is where the issues with NA's happen down the line.

01_financial-well-being/housing-value.qmd
```diff
@@ -0,0 +1,904 @@
...
+    count(year) |>
+    filter(year == 2022) |>
+    pull(n) == 3143
+)
+```
+
+Further examination shows that there are 5 counties in our metric data that
+are not in the county populations file. 
+```{r}
+#| label: find-extra-counties
```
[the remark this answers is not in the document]
malcalakovalski: @cdsolari how do you suggest we address this issue?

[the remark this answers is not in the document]
cdsolari: Thanks! yes, in other metrics we are having these all be N/A for CT because the county conversion is such a mess and we found out about it so close to our deadline. It is just for metrics that have 2022 data, and it will only be for 2022 for now. Prior years should still have CT values and should be ok. This is something we will deal with in our next round. Thanks! @malcalakovalski & @kmartinchek

01_financial-well-being/housing-value.qmd
malcalakovalski: I linked to the ipumsr vignette on [obtaining IPUMS data](https://tech.popdata.org/ipumsr/articles/ipums.html) since that's what I used it anyway!

01_financial-well-being/housing-value.qmd
malcalakovalski: Changing to statefip and countyfip fixed this. Thank you for flagging!!

01_financial-well-being/housing-value.qmd
malcalakovalski: Updated the comment accordingly!

01_financial-well-being/housing-value.qmd
malcalakovalski: Added a comment about this! I'm not clear as to why we do this, but PR isn't present in the ACS data I pull from IPUMS so this filtering is done to avoid NA's later on.

01_financial-well-being/housing-value.qmd
malcalakovalski: Thanks for catching this! The code indeed did not match what the text said, but the text was correct so I edited accordingly. I think this bug is part of the reason why there were missing values in the data quality flag.

01_financial-well-being/housing-value.qmd
malcalakovalski: Added a link to the specific notebook that describes this from the crosswalk folder!

01_financial-well-being/housing-value.qmd
malcalakovalski: One thing I noticed though is that the vast majority of flag values is 3. This is because currently `N` refers to homeowners in a county/place instead of population. This is pretty strict and causes more observations to have a data quality flag value of 3 compared to the geographic_allocation_quality variable in the puma county crosswalk file.

@jwalsh28 @cdsolari Should the population variable that determines quality be the total population or the population of homeowners?

01_financial-well-being/housing-value.qmd
kmartinchek: Perfect!

01_financial-well-being/housing-value.qmd
kmartinchek: Perfect!

01_financial-well-being/housing-value.qmd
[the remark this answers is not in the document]
kmartinchek: Thanks for double checking, this works now with the as_factor!

01_financial-well-being/housing-value.qmd
kmartinchek: Thank you for adding this note!

01_financial-well-being/housing-value.qmd
kmartinchek: Manu-- did you get a response from Claudia and John? Flagging that I will have to re-check this on a re-review once it has been confirmed.

01_financial-well-being/housing-value.qmd
kmartinchek: Thank you!

01_financial-well-being/housing-value.qmd
cdsolari: @malcalakovalski @kmartinchek I think the quality flag should be based on the total population. We expect that we will have people of color shut out of homeownership opportunities, and if they make up a large part of a community, but have a small share of the collective home value, that will be a strong quality finding. As opposed to if the group has a small share of the collective home values but they already make up such a small share of the community's population. So, please make quality based on the total population. Thanks!!!!!

01_financial-well-being/housing-value.qmd
kmartinchek: In reviewing the 2022 county file:
1. There are still values for 5 "old" CT counties. These should be removed.
2. Also, there should be observations added for all CT planning regions--only 6 are in the final dataset. 
3. For the metric value, if one part of the ratio is NA, should it be NA:0.8 (as it is written now) or just listed as NA?
4. For the data quality flag, the GitHub standards say that the quality flag should be "" if the metric is NA, but these are listed as NA in the final data, Is this right?
5. Even in recent year data files, should there be a year column? If yes, this should be added.

kmartinchek: In the subgroup files, I believe that the order of the subgroup within counties/places should be alphabetical.

kmartinchek:
I've completed a second review. Because IPUMS is down, I cannot run the full file, especially to verify changes to code to pull microdata that were made since my last review. I have requested some additional changes on CT counties in 2022, final formatting of files, data quality flags, and "missing" ratio values.

01_financial-well-being/housing-value.qmd
malcalakovalski: Will do, thank you!!

01_financial-well-being/housing-value.qmd
malcalakovalski: 1. I removed the 5 "old" CT counties and added the 3 missing planning regions in 2022.
2.  Claudia suggested leaving the data quality flag as currently written since this is something that can be handled in post-processing. 
3. Similarily, I confirmed with Aaron that since he read all files into R when post-processing, which is indifferent between “” and NA in .csvs, it's okay to leave it as NA instead of "".
4. I checked that `data/00_mobility-metrics_recent.csv` doesn't have a `year` column so it's okay that my recent file doesn't have one either!

kmartinchek:
The final adjustment is that there need to be rows for all 9 CT planning regions.

kmartinchek:
Before I approve, wanted to confirm the following and may need @cdsolari or @awunderground to weigh in based on treatment of other metrics:
1. Is the Valdez-Cordova Census Area in Alaska existing before 2020? It seems that 2014-2019 are NAs. If it isn't a county-equal during those periods, should it be dropped?
2. For places: it seems that South Fulton City is missing data for 2018, 2019, and 2020; although it was incorporated in 2017. Is that expected? 
3. When a place/county is missing for a specific year in the subgroup files, they just have 1 NA row (vs. one for overall, over + under 45). Is this the expected file format?

Everything else looks good!

cdsolari:
@kmartinchek @malcalakovalski Ideally, in the subgroup files, there should be placeholder rows for every subgroup, and not just "All." I'll let @awunderground weigh in on the importance of this over finishing though, given our timeline. For Alaska, Valdez-Cordova Census Area, Alaska (02-261), Split to form Chugach Census Area (02-063) and Copper River Census Area (02-066) effective January 02, 2019, as noted in this [documentation](https://www.census.gov/programs-surveys/geography/technical-documentation/county-changes.2010.html#list-tab-957819518). Careful that the population variable in the county-population.csv crosswalk has NA's for all of AK's counties for some reason. But, Valdez-Cordova should indeed cease to exist after 2019, and the other two counties should have replaced it thereafter.

kmartinchek:
Thanks @cdsolari for clarifying. I think the issue is that Valdez-Cordova CA is NA for 2014-2019 (even though it exists), before it disappears. Is that right?
image

cdsolari:
@kmartinchek Yes, that does seem weird that they are all NA for Valdez-Cordova. It could have been an issue in 2014 being NA if you used the population figure in the calculation, but it doesn't explain why the others are also NA. Maybe a merging issue? Makes sure the code is bysort year.

kmartinchek:
@cdsolari @malcalakovalski Great, thank you for clarifying! Seems like we should expect values for this county-equivalent for 2015-2018 (if not 2014 and 2019).

cdsolari:
@kmartinchek @malcalakovalski I think we should expect values for 2014-2019. If you don't have it for 2014, just make sure you don't use the population count from the crosswalk ;). It might be something funny with the merge. THANKS!

malcalakovalski:
@kmartinchek @cdsolari It looks like Valdez-Cordova CA isn't present in the puma to county crosswalk (`data/geographic-crosswalks/crosswalk_puma_to_county.csv`) . However, since it is present in the county population file, it shows up as NA in the final data.

awunderground:
This is exactly right. We needed a longitudinal file that represents geographies at a point in time since we need county drop downs that represent current counties. Valdez-Cordova, Alaska should be missing in those earlier years. 

So

awunderground:
My earlier comment was imprecise. @malcalakovalski, thank you for the precise language!

> It looks like Valdez-Cordova CA isn't present in the puma to county crosswalk (geographic-crosswalks/data/crosswalk_puma_to_county.csv). Therefore, we can't calculate metrics on it because there are no weights. However, since it is present in the county population file, it shows up as NA in the final data.

Your approach seems correct. It's perfectly fine for those earlier values to be missing. Just report the 3,142 counties with the `NA`s.

malcalakovalski:
**Note for posterity:** The code I wrote to add one missing row for each subgroup is quite hacky and hard to edit. In the future, I think it will be worth investing some time in refactoring it.

kmartinchek (approved):

PR #286, opened 2024-02-22
Iss209 -- transportation

tinatinc:
Transit trips & Transportation cost metrics - county & city, all & subgroup, for years 2015 & 2019. 
Not totally sure how to remove my earlier commits... the final code files are all uploaded at once in the most recent commit.


rpitingolo:
@tinatinc to make sure I don't miss anything, can you please confirm that these are 6 R scripts to be reviewed: 

- transit_cost_county.R
- transit_trips_cost_city.R
- transit_trips_cost_county_subgroups.R
- transit_trips_county.R
- transportation_index_race_ethnicity.R
- transportation_metrics_county.R

tinatinc:
> @tinatinc to make sure I don't miss anything, can you please confirm that these are 6 R scripts to be reviewed:
> 
> * transit_cost_county.R
> * transit_trips_cost_city.R
> * transit_trips_cost_county_subgroups.R
> * transit_trips_county.R
> * transportation_index_race_ethnicity.R
> * transportation_metrics_county.R

Hi, actually just the first four that you have listed here! 
The other two are old (I will check if I should move them somewhere). There should be one more called "transit_trips_cost_city_subgroup.R" which is missing! Not sure why, might be my mistake -- I will find and upload asap, but it shouldn't interfere it any of the others. 
Basically it's two metrics (transit trips, transit costs) from the same data source, at two geographies (city, county), and two levels (overall, subgroup).

awunderground:
You can delete the old files. The old files are stored in the commit history.

rpitingolo:
@tinatinc thanks. Please go ahead and commit on this branch so I can refresh with the correct list of scripts.

06_neighborhoods/Transportation/transit_trips_county.R
```diff
@@ -0,0 +1,203 @@
...
+
+###################################################################
+
+# 1. Import all the files (and/or combine into one file) with only the relevant variables and years
+
+# FOR COUNTIES - import the raw data files
+# Unlike for Places and Tracts, the CNT website allows you to download raw data for all counties at once
+# so we just import the all-county file for the two available years; 2015 and 2019
+# 2015
+transport_county_2015 <- read_csv("C:/Users/USERNAME/Box/Lab/Projects/Gates Upward Mobility Framework/Outreach and Tools/Data/Metrics_2024_round/Transportation/htaindex2015_data_counties.csv")
```
rpitingolo, line 35: Ideally code would download the CSV from the website into a temp folder and then read it in. It looks like the data is behind a login screen so this may not be possible but is worth investigating as it is good practice.

Otherwise the issue with Box Drive is that the path is permission specific. For example I have permission to  the "Data" folder but not any of the root folders. If we use Box Drive I suggest creating a single line of code at the top where the person running the code can put their file path, then use past0 in the read_csvs.

06_neighborhoods/Transportation/transit_trips_county.R
```diff
@@ -0,0 +1,203 @@
...
+  )
+
+# keep only variables of interest
+transit_trips_county_2015 <- transport_county_2015 %>%
+  select(state, county, blkgrps, population, households, transit_trips_80ami)
+
+
+#2019
+transport_county_2019 <- read_csv("C:/Users/USERNAME/Box/Lab/Projects/Gates Upward Mobility Framework/Outreach and Tools/Data/Metrics_2024_round/Transportation/htaindex2019_data_counties.csv")
+transport_county_2019 <- transport_county_2019 %>%
```
rpitingolo, line 51: A lot of code in this script (and the others) is harded coded twice. Converting it to a function will reduce the chance of error and make adding new years of data easier if we need to add more in the future.

06_neighborhoods/Transportation/transit_trips_county.R
```diff
@@ -0,0 +1,203 @@
...
+    county = substr(county, start = 4, stop = 6)
+  )
+
+# keep only variables of interest
+transit_trips_county_2019 <- transport_county_2019 %>%
+  select(state, county, blkgrps, population, households, transit_trips_80ami)
+
+# Compare to our official county file to make sure we have all counties accounted for
+# Bring in the official county file
+counties <- read_csv("C:/Users/USERNAME/Documents/GitHub/mobility-from-poverty/geographic-crosswalks/data/county-populations.csv")
```
rpitingolo, line 67: Since we have a .rproj that we ask people to use, this should be soft coded instead of hard coded.

`counties <- read_csv("geographic-crosswalks/data/county-populations.csv")`

06_neighborhoods/Transportation/transit_trips_county.R
```diff
@@ -0,0 +1,203 @@
...
+
+
+# County-Level Transit Trips 2019
+ggplot(transit_trips_county_2019, aes(x=transit_trips_80ami)) + geom_histogram(binwidth=5) + labs(y="number of counties", x="Annual Transit Trips for the Regional Moderate Income Household, 2019")
+# looks similar, checks out
+# look at summary stats
+summary(transit_trips_county_2019$transit_trips_80ami)  
+# examine outliers
+transit_trips_county_2019_outliers <- transit_trips_county_2019 %>% 
+  filter(transit_trips_80ami>250)
```
rpitingolo, line 118: If you are creating a dataframe for manual outlier checking it is helpful to sort it largest to smallest.

06_neighborhoods/Transportation/transit_trips_county.R
```diff
@@ -0,0 +1,203 @@
...
+  print(transit_trips_county_2019[missing_indices, , drop = FALSE])
+} else {
+  cat("No missing values for transit_trips_county_2019\n")
+}
+# No missing values for 2019
+
+#####################################################################
+
+# Create national percentile ranking for 'transit_trips_80ami'
+transit_trips_county_2015$rank <- rank(transit_trips_county_2015$transit_trips_80ami)
```
rpitingolo, line 142: This code is a mix of base R and tidyverse syntax. It's good practice to stick with one. Since you already have some tidyverse I would suggest sticking with that. This section of code can be re-written more efficient and even more so if it's wrapped in a function. 

```
transit_trips_county_2015 <- transit_trips_county_2015 %>%
  mutate(rank = rank(transit_trips_80ami),
         percentile_rank = (rank - 1) / (n() - 1) * 100,
         percentile_rank = round(percentile_rank, 2)) %>%
  rename(index_transit_trips = percentile_rank)
```

06_neighborhoods/Transportation/transit_trips_county.R
```diff
@@ -0,0 +1,203 @@
...
+
+
+# Combine the two years into one overall file
+transit_trips_county <- rbind(transit_trips_county_2015, transit_trips_county_2019)
+
+
+# Keep variables of interest and order them appropriately
+transit_trips_county <- transit_trips_county %>%
+  select(year, state, county, index_transit_trips, index_transit_trips_quality) %>%
+  select(year, state, county, index_transit_trips, index_transit_trips_quality)
```
rpitingolo, line 196: Second line "select" is redundant. Is this supposed to be "arrange"?

06_neighborhoods/Transportation/transit_cost_county.R
```diff
@@ -0,0 +1,189 @@
...
+  # check that all values are non-negative & count missing values
+  # examine outliers
+  
+
+  # County-Level Transportation Cost 2015
+ggplot(transportation_cost_county_2015, aes(x=t_80ami)) + geom_histogram(binwidth=10) + labs(y="number of counties", x="Annual Transit Cost for the Regional Moderate Income Household, 2015")
+# look at summary stats
+summary(transportation_cost_county_2015$t_80ami)  
+# examine outliers
+transportation_cost_county_2015_outliers <- transportation_cost_county_2015 %>%
```
rpitingolo, line 91: How did you decide on 100 for outliers? Not saying it's wrong just not sure what it is based on (a percentile?)

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,283 @@
+
+
+###################################################################
+
+# Transit Metrics: Transportation Costs & Trips - CITY LEVEL
```
rpitingolo, line 5: This script does not load any packages. At a minimum I believe it needs at least the tidyverse package to run.

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,283 @@
...
+
+# 2015
+# bring in all the downloaded CSVs (state-level tracts) & combine them into one nation-wide file for tracts
+# these files can be found in the Urban Box folder -- replace USERNAME
+tracts15files <- list.files(path="C:/Users/USERNAME/Box/Lab/Projects/Gates Upward Mobility Framework/Outreach and Tools/Data/Metrics_2024_round/Transportation/2015_tract",
+                            pattern="*.csv")
+print(tracts15files)
+tractpath15 = file.path("C:/Users/USERNAME/Box/Lab/Projects/Gates Upward Mobility Framework/Outreach and Tools/Data/Metrics_2024_round/Transportation/2015_tract",tracts15files)
+print(tractpath15)
+transport_tracts_2015 <- do.call("rbind",lapply(tractpath15,FUN=function(files){ read.csv(files)}))
```
rpitingolo, line 32: Can also use tidyverse syntax to read in the files.

`transport_tracts_2015 <- map_df(tractpath15, read_csv)`

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,283 @@
...
+# For when we want to: Limit to the Census Places we want 
+# first, bring in the places crosswalk (place-populations.csv)
+places <- read_csv("geographic-crosswalks/data/place-populations.csv")
+# keep only the most relevant year
+places <- places %>%
+  filter(year == 2019)
+
+
+# TRANSPORTATION COST 2015
+transit_cost_city_2015 <- left_join(transit_cost_tracts_2015, tract_place, by=c("state", "tract"))
```
rpitingolo, line 106: 🚨🚨🚨 I believe this merge is incorrect. 

My understanding is that you are trying to merge two datasets BY TRACT. In this case you are using a state ID and a tract ID. But tracts are unique to counties, not states, so you are potentially creating many duplicates here. The resulting dataset when I run it has 446,846 rows, whereas a proper left join using unique tracts should result in a dataframe that is 123,633 (assuming tracts are the same in both files). 

Because of this, I believe everything below that uses transit_cost_city_2015 is incorrect.

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,283 @@
...
+# keep only the variables we will need & replace NA qual vars where there is NA metric value
+transit_cost_city_2015 <- transit_cost_city_2015 %>% 
+  mutate(year=2015)%>%
+  select(year, state, place, index_transportation_cost, index_transportation_cost_quality)%>%
+  dplyr::mutate(index_transportation_cost_quality = ifelse(is.na(index_transportation_cost), NA, index_transportation_cost_quality))
+
+
+
+# TRANSPORTATION COST 2019
+transit_cost_city_2019 <- left_join(transit_cost_tracts_2019, tract_place, by=c("state", "tract"))
```
rpitingolo, line 133: Same 🚨🚨🚨 issue as above. Instead of joining by two variables a better practice would be to merge based on a unique geo ID in the format of: ssccctttttt. 

I believe everything below that uses transit_cost_city_2019 is incorrect.

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,283 @@
...
+# keep only the variables we will need & replace NA qual vars where there is NA metric value
+transit_cost_city_2019 <- transit_cost_city_2019 %>% 
+  mutate(year=2019)%>%
+  select(year, state, place, index_transportation_cost, index_transportation_cost_quality)%>%
+  dplyr::mutate(index_transportation_cost_quality = ifelse(is.na(index_transportation_cost), NA, index_transportation_cost_quality))
+
+
+
+# TRANSIT TRIPS 2015
+transit_trips_city_2015 <- left_join(transit_trips_tracts_2015, tract_place, by=c("state", "tract"))
```
rpitingolo, line 155: Same merge issue.

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,283 @@
...
+  mutate(
+    rank = rank(index_transit_trips),
+    percentile_rank = ((rank - 1) / (n() - 1)) * 100,
+    index_transit_trips = round(percentile_rank, 2)
+  )%>%
+  dplyr::mutate(index_transit_trips_quality = ifelse(dq / n > 0.5, 1, ifelse(dq / n <= 0.5, 2, NA)))
+
+
+# left join with places file to get rid of irrelevant places data
+transit_trips_city_2015 <- left_join(places, transit_trips_city_2015, by=c("state","place"))
```
rpitingolo, line 178: Same merge issue.

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,645 @@
+############################################################
+
+# SUBGROUP FILE
+
+# Process for SUBGROUP-level:
```
rpitingolo, line 5: This script does not load any packages. I believe you need at a minimum tidyverse and tidycensus.

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,645 @@
...
+cat("Number of missing values for transit_trips_80ami:", missing_count, "\n")
+# 831 missing values - aligns with cost data for the same year, good
+
+############################################################################
+
+# For each tract, ID the race category
+# Calculate the percentage of total population for White & POC for each tract
+cost_data_2015 <- cost_data_2015 %>%
+  mutate(perc_white = if_else(population > 0, white_population / total_population, 0), 
+         perc_POC = if_else(population > 0, ((black_population + asian_population + aian_population + nhpi_population + other_population) / total_population), 0))
```
rpitingolo, line 225: Are the component variables always zero if  they are zero? Be careful to account for potential missingness when using this syntax to do addition.

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,645 @@
...
+         subgroup = "All")
+
+trips_all_19 <- trips_data_2019 %>%
+  group_by(state, county) %>%
+  summarize(index_transit_trips = weighted.mean(x = transit_trips_80ami, w = households, na.rm = TRUE)) %>%
+  mutate(subgroup_type = "race-ethnicity",
+         subgroup = "All")
+
+# Now make sure to change the trips to national percentile ranks
+trips_all_15$rank <- rank(trips_all_15$index_transit_trips)
```
rpitingolo, line 320: When I run this code it appears NAs are getting ranked. Is that right? It doesn't seem right to me.

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,645 @@
...
+# keep only what we need
+trips_all_15 <- trips_all_15 %>%
+  select(state, county, index_transit_trips, subgroup_type, subgroup)
+trips_all_19 <- trips_all_19 %>%
+  select(state, county, index_transit_trips, subgroup_type, subgroup)
+
+
+# 3b. Collapse accordingly -- to counties and race categories, weighting the measure by HH count per tract
+
+cost_by_race_15 <- cost_data_2015 %>%
```
rpitingolo, line 337: A lot of NAs are created here. I'm not sure if that is supposed to happen.

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,645 @@
...
+  rename(subgroup = race_category)
+
+# remove dummy df
+rm(county_expander)
+
+
+
+# CHANGE THE TRIPS VALUES TO NATIONAL PERCENTILE RANKS FOR EACH
+# For 2015 data
+trips_by_race_15 <- trips_by_race_15 %>%
```
rpitingolo, line 386: Maybe I am missing something but a lot of NAs are getting ranked here too?

rpitingolo (changes requested):
@tinatinc I've finished reviewing. There are potentially some critical merge errors on the city script so please take a look at that first. Thanks!

06_neighborhoods/Transportation/transit_trips_county.R
tinatinc: Agreed - I don't have time to do this this round but will make a note for the team

06_neighborhoods/Transportation/transit_trips_cost_city.R
tinatinc: Spot on -- new df obs count is correct, thank you for catching!

06_neighborhoods/Transportation/transit_trips_cost_city.R
tinatinc: Yes -- new df count is 123825 obs

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
tinatinc: Good call - I've updated these snippets accordingly so that if these columns have missing values for a particular observation, the corresponding perc_white or perc_POC value for that observation is set to NA. e.g., calculations are performed only when all data are available

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
tinatinc: Updated to calculate ranks for only non-NA values (rows 370-380)

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
tinatinc: Added in a check for this -- looks like it's OK

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
tinatinc: Yes, thank you for catching! Updated to exclude NA values in the ranking

06_neighborhoods/Transportation/transit_trips_cost_city.R
rpitingolo: Merge looks correct now.

06_neighborhoods/Transportation/transit_trips_cost_city.R
rpitingolo: Merge looks correct now.

06_neighborhoods/Transportation/transit_trips_cost_city.R
rpitingolo: Merge looks correct now.

06_neighborhoods/Transportation/transit_trips_cost_city.R
rpitingolo: Merge looks correct now.

06_neighborhoods/Transportation/transit_trips_county.R
rpitingolo: OK, fair. I am going to mark this comment as resolved.

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,292 @@
...
+
+# TRANSPORTATION COST 2015
+transit_cost_city_2015 <- left_join(transit_cost_tracts_2015, tract_place, by=c("state", "county", "tract"))
+# collapse to places and also create data quality marker
+# data quality can be 1 when most of the tracts that fall in the place (e.g., >50% of the tracts) 
+# have most of their area falling in the place (e.g., >50% of the tract's area is in the place)
+# otherwise, data quality is 2
+transit_cost_city_2015 <- transit_cost_city_2015 %>% 
+  dplyr::group_by(state, place) %>% 
+  dplyr::summarize(index_transportation_cost = round(mean(t_80ami), 2),
```
rpitingolo, line 122: If I understand this code correctly, you are trying to take a mean of the variable `t_80ami` BY place by averaging the value of each component tract together. If this this case doesn't this need to be a weighted mean using afact? 

Same comment applies to 2019 cost.

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,292 @@
...
+                   n = n(),
+                   dq = sum(afact > 0.5, na.rm = TRUE)
+  )%>%
+  dplyr::mutate(index_transportation_cost_quality = ifelse(dq / n > 0.5, 1, ifelse(dq / n <= 0.5, 2, NA)))
```
rpitingolo, line 126: You are using `afact` here to create the quality score, which I think is OK if that's how you want to define it, but I still think the mean itself needs to be weighted by `afact`.

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,292 @@
...
+
+# TRANSIT TRIPS 2015
+transit_trips_city_2015 <- left_join(transit_trips_tracts_2015, tract_place, by=c("state", "county","tract"))
+# collapse to places and also create data quality marker
+# data quality can be 1 when most of the tracts that fall in the place (e.g., >50% of the tracts) 
+# have most of their area falling in the place (e.g., >50% of the tract's area is in the place)
+# otherwise, data quality is 2
+transit_trips_city_2015 <- transit_trips_city_2015 %>% 
+  dplyr::group_by(state, place) %>% 
+  dplyr::summarize(index_transit_trips = sum(transit_trips_80ami),
```
rpitingolo, line 171: Should this be a weighted sum by `afact`? Same comment applies to 2019 trips.

rpitingolo (changes requested):

06_neighborhoods/Transportation/transit_cost_county.R
```diff
@@ -0,0 +1,189 @@
...
+###################################################################
+
+# 1. Import all the files (and/or combine into one file) with only the relevant variables and years
+
+# FOR COUNTIES - import the raw data files
+# Unlike for Places and Tracts, the CNT website allows you to download raw data for all counties at once
+# so we just import the all-county file for the two available years; 2015 and 2019
+# 2015
+# raw data lives in the Urban Box folder (replace USERNAME)
+transport_county_2015 <- read_csv("C:/Users/USERNAME/Box/Lab/Projects/Gates Upward Mobility Framework/Outreach and Tools/Data/Metrics_2024_round/Transportation/htaindex2015_data_counties.csv")
```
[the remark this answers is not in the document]
tinatinc: The automation did not work because of the ZIP nature of each file and the need for concatenation (maybe someone can try again in the future), but data download description will be added to the documentation

06_neighborhoods/Transportation/transit_trips_cost_city.R
tinatinc: Yes! Thank you -- adjusted both

06_neighborhoods/Transportation/transit_trips_cost_city.R
tinatinc: Agreed, adjusted

06_neighborhoods/Transportation/transit_trips_cost_city.R
tinatinc: Yes, thank you! added

06_neighborhoods/Transportation/transit_trips_county.R
```diff
@@ -0,0 +1,201 @@
...
+###################################################################
+
+
+# Libraries you'll need
+library(tidyr)
+library(dplyr)
+library(readr)
+library(ggplot2)
+library(tidyverse)
+library(purrr)
```
rpitingolo, line 21: Tidycensus is still missing from the list of packages

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
rpitingolo: This still seems conceptually weird to me. NAs are created in situations where a tract essentially had no population. So a county that has all populated tracts won't have any NA. But a county with some non-populated tracts would. I don't quite know what the NA does on cost_by_race_yy. 

Maybe it doesn't matter since they get dropped off when you merge on the county expander.

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,729 @@
...
+  summarize(transit_cost = weighted.mean(x = t_80ami, w = households, na.rm = TRUE))
+
+na_count <- sum(is.na(cost_by_race_15$transit_cost))
+# went from 810 missings to 196 and the values look consistent
+
+cost_by_race_19 <- cost_data_2019 %>%
+  group_by(state, county, race_category) %>%
+  summarize(transit_cost = weighted.mean(x = t_80ami, w = households, na.rm = TRUE))
+
+trips_by_race_15 <- trips_data_2015 %>%
```
rpitingolo, line 412: This is county-level weighted mean. Is that correct or should it be weighted sum?

rpitingolo (changes requested):

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,292 @@
...
+
+
+###################################################################
+
+# 5. Export final files
+
+
+# Combine the two years into one overall files for both variables
+transit_cost_city <- rbind(transit_cost_city_2015, transit_cost_city_2019)
+transit_trips_city <- rbind(transit_trips_city_2015, transit_trips_city_2019)
```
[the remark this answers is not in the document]
tinatinc: Yes, that's right -- because the purpose of creating the ranking is to contextualize it nationally, I felt it made more sense to rank with all the places, then limit to the ones we highlight. I can run this thinking past Claudia if it's worrisome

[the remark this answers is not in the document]
tinatinc: Spoke with Claudia about this -- we feel comfortable with this decision, but we will consult with the team to beef up language explaining this unique situation. Thanks for flagging

06_neighborhoods/Transportation/transit_trips_cost_city.R
```diff
@@ -0,0 +1,292 @@
...
+# data quality can be 1 when most of the tracts that fall in the place (e.g., >50% of the tracts) 
+# have most of their area falling in the place (e.g., >50% of the tract's area is in the place)
+# otherwise, data quality is 2
+transit_cost_city_2015 <- transit_cost_city_2015 %>% 
+  dplyr::group_by(state, place) %>% 
+  dplyr::summarize(index_transportation_cost = round(weighted.mean(t_80ami, w = afact, na.rm = TRUE), 2),
+                   n = n(),
+                   dq = sum(afact > 0.5, na.rm = TRUE)
+  )%>%
+  dplyr::mutate(index_transportation_cost_quality = ifelse(dq / n > 0.5, 1, ifelse(dq / n <= 0.5, 2, NA)))
```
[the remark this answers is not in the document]
tinatinc: Just spoke with Claudia -- this is aligned with the cutoff we use for air quality (which does a similar collapse). Updating to tweak it to match exactly

06_neighborhoods/Transportation/transit_trips_county.R
```diff
@@ -0,0 +1,201 @@
...
+# 3. Data Quality marker
+
+# Determine data quality cutoffs based on number of observations (all at the HH level for these values)
+summary(transit_trips_county_2015$households) 
+summary(transit_trips_county_2019$households) 
+
+# we use a 30 HH cutoff for Data Quality 3 for the ACS variables, so for the sake of consistency, since none of 
+# these are less than 30 (all minimum values are at least 30 HHs), Data Quality can be 1 for all these observations
+transit_trips_county_2015 <- transit_trips_county_2015 %>% 
+  mutate(index_transit_trips_quality = 1)
```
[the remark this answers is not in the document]
tinatinc: Agreed -- updating accordingly

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
tinatinc: For this one, t_80ami is actually a percentage value, not a count like the other variable in the dataset (trips), so it makes sense to mean not sum

06_neighborhoods/Transportation/transit_trips_county.R
tinatinc: Added

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,729 @@
...
+  filter(index_transportation_cost>100) 
+# Nothing unexpected
+
+# Use stopifnot to check if all values are non-negative
+stopifnot(min(cost_by_race_19$index_transportation_cost, na.rm = TRUE) >= 0)
+# Good to go
+
+
+
+# Trips 2015
```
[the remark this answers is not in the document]
tinatinc: This is normal/to be expected, to my understanding. Most counties are rural, or have little to no public transit infrastructure, so the distribution will skew low (minimal transit trips) when looking at the whole country (as opposed to, for example, the top 500 cities)

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,729 @@
...
+
+# Use stopifnot to check if all values are non-negative
+stopifnot(min(trips_by_race_19$index_transit_trips, na.rm = TRUE) >= 0)
+# Good to go
+
+
+###################################################################
+
+# 5. Data Quality marker
+# none of these values are coming from estimates with less than 30 observations
```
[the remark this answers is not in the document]
tinatinc: Agreed - updated accordingly using households as the threshold cutoff

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,729 @@
...
+
+cost_all_19 <- cost_data_2019 %>%
+  group_by(state, county) %>%
+  summarize(index_transportation_cost = weighted.mean(x = t_80ami, w = households, na.rm = TRUE)) %>%
+  mutate(subgroup_type = "race-ethnicity",
+         subgroup = "All")
+
+trips_all_15 <- trips_data_2015 %>%
+  group_by(state, county) %>%
+  summarize(index_transit_trips = weighted.mean(x = transit_trips_80ami, w = households, na.rm = TRUE)) %>%
```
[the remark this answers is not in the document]
rpitingolo: Just to make sure I understand this data - is the transit_trips tract-level data the number of trips per household or the total number of trips in the tract? If the former I think the updated code is OK. If the latter then it seems like the weight needs to be the percentage of households in the county? 

Though maybe it doesn't matter if when you create the ranks the values of the rank are the same regardless.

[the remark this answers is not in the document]
tinatinc: Hi yes -- it's the number of trips per household ("Annual Transit Trips for the Regional Moderate Household"), here is the data dictionary for the source data if interested: https://htaindex.cnt.org/download/data-dictionary.php

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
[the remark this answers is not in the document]
rpitingolo: OK, I am still struggling to understand how to interpret this distribution but if it's confirmed correct I will mark this as resolved.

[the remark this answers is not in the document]
tinatinc: OK, cool! @cdsolari do you want to take a peek/could you confirm this tracks with intuition here?

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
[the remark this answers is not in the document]
rpitingolo: Great, thanks!

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
rpitingolo: I am going to resolve this one because I think it's OK once you merge on the expander.

06_neighborhoods/Transportation/transit_trips_cost_city_subgroups.R
```diff
@@ -0,0 +1,789 @@
...
+
+# Keep only our cities of interest (486 of them from the places crosswalk)
+places <- read_csv("geographic-crosswalks/data/place-populations.csv")
+# keep the most recent year
+places <- places %>%
+  filter(year == 2019) %>%
+  select(state, place)
+
+# left join to limit to places of interest
+transit_trips_subgroup_city <- left_join(places, transit_trips_subgroup_city, by = c("state", "place"))
```
rpitingolo, line 776: The final file has 3,867 observations. Since each place should have 4 rows times 2 years the total observations should be 3,888 (assuming 486 valid places). I would suggest running a frequency to see which place has less than 8 to diagnose what is going wrong.

rpitingolo (changes requested):

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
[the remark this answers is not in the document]
tinatinc: Just spoke with Claudia -- we think this tracks based on the choices we've made (when to rank, etc.), but made a note that we will go out of our way to explain context and interpretation in the support tools/table notes

06_neighborhoods/Transportation/transit_trips_cost_city_subgroups.R
tinatinc: Thank you for catching - these were missing data that were not being accounted for in the final output. Updating accordingly so the output files account for every valid place & subgroup group

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,740 @@
...
+transit_trips_tracts_2019$GEOID <- paste0(transit_trips_tracts_2019$state, transit_trips_tracts_2019$county, transit_trips_tracts_2019$tract)
+
+
+# merge population data with transit data files (4 of them)
+############################################################################
+# FIRST ONE - Cost in 2015
+############################################################################
+cost_data_2015 <- left_join(acs_tract_pop15, transit_cost_tracts_2015, by = "GEOID")
+#Test with anti_join to make sure it worked properly
+stopifnot(
```
rpitingolo, line 177: These `stopifnot` sections are preventing the script from running all the way through top to bottom. They are useful for testing / debugging but I believe we are supposed to have final code that runs with no errors or issues. So we need to either switch this to some kind of warning output or remove it since we trust the code is working as intended. 

Same comment applies to the other scripts where  `stopifnot` is used.

rpitingolo (changes requested):

06_neighborhoods/Transportation/transit_trips_cost_city_subgroups.R
tinatinc: Yes -- two place in Georgia (place 49008 and 72122) were missing in the source data entirely, and place 37692 in California was dropping it's NA obs in the subgroup expansion

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
tinatinc: That makes sense -- I've commented these out of the scripts since they are process code not necessary to the output, but am leaving them in as comments to show documentation, missing counts, etc.

06_neighborhoods/Transportation/transit_trips_cost_city_subgroups.R
rpitingolo: Nice. Thanks.

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
```diff
@@ -0,0 +1,740 @@
...
+#missing_count <- sum(is.na(trips_data_2015$transit_trips_80ami))
+#cat("Number of missing values for transit_trips_80ami:", missing_count, "\n")
+# 810 missing values - aligns with cost data for the same year, good
+
+############################################################################
+# FOURTH ONE - Trips in 2019
+############################################################################
+trips_data_2019 <- left_join(acs_tract_pop19, transit_trips_tracts_2019, by = "GEOID")
+#Test with anti_join to make sure it worked properly
+stopifnot(
```
rpitingolo, line 219: `stopifnot` code is still active

06_neighborhoods/Transportation/transit_trips_cost_city_subgroups.R
```diff
@@ -0,0 +1,811 @@
...
+#missing_count <- sum(is.na(cost_data_2019$t_80ami))
+#cat("Number of missing values for t_80ami:", missing_count, "\n")
+# 831 missing values
+
+############################################################################
+# THIRD ONE - Trips in 2015
+############################################################################
+trips_data_2015 <- left_join(acs_tract_pop15, transit_trips_tracts_2015, by = "GEOID")
+#Test with anti_join to make sure it worked properly
+stopifnot(
```
rpitingolo, line 205: `stopifnot` code is still active

rpitingolo (changes requested):

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
rpitingolo: There are still places where the code is active. Please double check that scripts run through start to finish. Thanks!

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
tinatinc: removed

06_neighborhoods/Transportation/transit_trips_cost_city_subgroups.R
tinatinc: removed

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
tinatinc: ahhh!! Sorry!! Removed  & updated

06_neighborhoods/Transportation/transit_trips_cost_county_subgroups.R
rpitingolo: This one is still not running all the way through

tinatinc: Shoot.. running fine on my end, so maybe a package issue? Where does it error for you?

rpitingolo: Line 191 seems to be the problem

tinatinc: Darn!! I still missed one! Thanks Rob, should be good now

rpitingolo: Were you able to get it to run all the way through? I am now getting errors on the `get_acs` section (an API issue not a code issue) so I'm not able to confirm the updated section works yet.

tinatinc: Yes, runs through on my end

rpitingolo (approved):

PR #289, opened 2024-02-22
Iss216 -- high-poverty schools

JCarterUI:
Share of students attending high-poverty schools, by student race or ethnicity ready for review


JCarterUI:
I'm done with this now and it's ready for review

dgon32 (changes requested):
code for both files stops because of references to Box folders. In addition to comments throughout, update to remove box folder references.

cdsolari (changes requested):
The big thing is that the race/ethnic groups for this metric are not designed as a subgroup. We have no "all" value. The metric is "share of students attending high-poverty schools, by student race/ethnicity" So it require a variable for each group of black, Hispanic and white for the overall. There's no subgroup file for that reason. And there's no All value. It is a diversity metric. Can you please reshape the data back to wide and let me know if you have questions.

cdsolari (changes requested):
This all looks good! Only question is if those intermediate files are properly in git ignore. Please do a quick check on that. Thanks!

cdsolari (approved):
Everything looks great!

PR #294, opened 2024-02-28
Iss220 -- broadband access

malcalakovalski:
This metric shows the ratio of households (or people in households) with the presence of a computer and a 
type of broadband internet subscription. In particular, the unit of analysis for the overall and income subgroup is households, whereas the unit of analysis for the race/ethnicity subgroup is people in households. 

This metric is calculated using `digital_access.qmd` in the 08_education folder, which pulls 1-year ACS data for the overall metric and 5-year ACS data for the subgroups. 

**Note:** While this branch is ready for review, it is still missing confidence interval calculations for the metric. Therefore, there aren't columns with `_lb` and `_ub` suffixes in the final data yet. @cdsolari will provide guidance on how to proceed.


malcalakovalski:
Confidence intervals and updated data quality flags have been added

08_education/digital_access.qmd
```diff
@@ -0,0 +1,939 @@
...
+
+We are using the 1-yr ACS for the communities overall (not the subgroups), from
+2016 to 2022. This will be based on table B28003.
+
+```{r}
+# "Population" variables refer to households/people in households in a
+# given category.
+
+# "Digital" variables (those with a "digital" prefix) refer to households/people in households in a given category that have a computer with a broadband internet subscription.
+
```
tinatinc, line 108: Not a need, this is just a preference thing for clarity - I like to add the census codes & their names as a comment for reference before introducing in the code. Like: 
B28003 - Presence of a Computer and Type of Internet Subsciption in Household
_001 - Total
_004 With a Broadband Internet Subscription

08_education/digital_access.qmd
```diff
@@ -0,0 +1,939 @@
...
+      other_digital = asian_digital + aian_digital + nhpi_digital + other_digital + two_or_more_digital
+    ) |>
+    mutate(
+      moe_other_digital = sqrt(
+        moe_aian_digital^2 +
+          moe_asian_digital^2 +
+          moe_nhpi_digital^2 +
+          moe_other_digital^2 +
+          moe_two_or_more_digital^2
+      )
```
tinatinc, line 218: I don't dispute this square root method is valid, but would maybe add a line of explanation for this MOE aggregation process

tinatinc: Ah, I see you link out to something later in the code -- if that's the same rationale, maybe just point to it

08_education/digital_access.qmd
```diff
@@ -0,0 +1,939 @@
...
+map_dbl(county_income_digital_access, ~ sum(is.nan(.x)))
+map_dbl(place_income_digital_access, ~ sum(is.nan(.x)))
+```
+
+# Create data quality flags
+
+We only look at the CVs for the numerator, which is less precise than the
+denominator. We use 0.1, 0.2, and 0.5 as thresholds because the CV only accounts
+for sampling error in the numerator. This is stricter than for some other
+metrics.
```
tinatinc, line 466: OK but the numbers themselves are determined based on distribution?

08_education/digital_access.qmd
```diff
@@ -0,0 +1,939 @@
...
+place_overall_digital_access |> get_summary_stats()
+
+county_raceeth_digital_access |> get_summary_stats()
+
+place_raceeth_digital_access |> get_summary_stats()
+
+county_income_digital_access |> get_summary_stats()
+
+place_income_digital_access |> get_summary_stats()
+```
```
tinatinc, line 887: could we add a comment here re: how many get suppressed when we do this?

documentation/overview.md
```diff
@@ -1002,20 +1002,30 @@ Please note that the denominator we use is the living wage for a single full-tim
...
 ### Overview
 
-* **Analyst & Programmer:** Tina Chelidze
-* **Year(s):** 2021
-* **Final data name(s):** `digital_access_county_2021.csv` and `digital_access_city_2021.csv'
-* **Data Source(s):** Census Bureau's American Community Survey 
-* **Notes:**
-* **Data Quality Index:** `1` means the observation count informing the estimate is greater than or equal to 30. The quality index is a `2` otherwise.
+* **Analyst & Programmer:** Tina Chelidze and Manuel Alcalá Kovalski
+* **Year(s):** 2016, 2017, 2018, 2019, 2021, 2022
```
tinatinc, line 1006: it looks like your output files have 2020 as well?

tinatinc:
OK everything else in the code looks good! Except you have an issue in your subgroup output files:
for both "digital_access_county_subgroup_all.csv" and "digital_access_city_subgroup_all.csv", the subgroup_type variable for the income subgroups reads as "NA" -- should read as "income". Also you should have a third observation paired with each income breakdown (like you do with race) that has the "all" value. 

I also made a note in the documentation that you seem to have included 2020 data, so we should note that there.

08_education/digital_access.qmd
malcalakovalski: Moved this up so folks can see it the first time I do this aggregation!

08_education/digital_access.qmd
malcalakovalski: This comment was actually outdated so I removed it. However, I updated the text to mention that we are using the same cutoffs for 1-year vs 5-year ACS

08_education/digital_access.qmd
malcalakovalski: I found that no observations were actually supressed because there aren't any missing quality flag values when the metric
value is not missing! Added a comment saying this.

documentation/overview.md
malcalakovalski: Thanks for catching this! It looks like it was included in the output but all values were NA. So I now filter 2020 before exporting!

tinatinc:
Hey @malcalakovalski - re-reviewed and things look good. I can't complete review until census API is running again, so waiting on that. 
Just noting -- a lot of missings in the non-subgroup files, and I'm not seeing a count of missings for each output file in the code. Am I right that, for example, for the non-subgroup overall county-level data, we only have this metric for like 25% of counties?

malcalakovalski:
Hey @tinatinc. It looks like the reason we have a large number of missing values in the non-subgroup county files is that we switched from ACS 5-year to ACS 1-year. I found that there were only 4,969 observations right after pulling the county overall ACS 1 year data from tidycensus. You can check this by running the chunk labelled "pull-overall-communities". This translates to between 820 and 837 observations per year. 

By comparison, running:

```
tidycensus::get_acs(geography = "county",
                    variables = "B28003_004",
                    year = 2021,
                    survey = "acs5")
```

gives me 3,221 observations. 

So the good news is nothing in the code is inadvertadely creating these NA's or removing observations we want. However, we may or may not want to revisit using the 1 year ACS given the significant increase in missing values. @cdsolari how should we proceed?

tinatinc:
> Hey @tinatinc. It looks like the reason we have a large number of missing values in the non-subgroup county files is that we switched from ACS 5-year to ACS 1-year. I found that there were only 4,969 observations right after pulling the county overall ACS 1 year data from tidycensus. You can check this by running the chunk labelled "pull-overall-communities". This translates to between 820 and 837 observations per year.
> 
> By comparison, running:
> 
> ```
> tidycensus::get_acs(geography = "county",
>                     variables = "B28003_004",
>                     year = 2021,
>                     survey = "acs5")
> ```
> 
> gives me 3,221 observations.
> 
> So the good news is nothing in the code is inadvertadely creating these NA's or removing observations we want. However, we may or may not want to revisit using the 1 year ACS given the significant increase in missing values. @cdsolari how should we proceed?

Thank you for checking on this Manu!! OK, I spoke with Claudia about this -- because the coverage is so low (like 1/4 of all counties), we think it's probably better to switch back to 5-year estimates for the overall values as well. Is this something you can do?

malcalakovalski:
> > Hey @tinatinc. It looks like the reason we have a large number of missing values in the non-subgroup county files is that we switched from ACS 5-year to ACS 1-year. I found that there were only 4,969 observations right after pulling the county overall ACS 1 year data from tidycensus. You can check this by running the chunk labelled "pull-overall-communities". This translates to between 820 and 837 observations per year.
> > By comparison, running:
> > ```
> > tidycensus::get_acs(geography = "county",
> >                     variables = "B28003_004",
> >                     year = 2021,
> >                     survey = "acs5")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > gives me 3,221 observations.
> > So the good news is nothing in the code is inadvertadely creating these NA's or removing observations we want. However, we may or may not want to revisit using the 1 year ACS given the significant increase in missing values. @cdsolari how should we proceed?
> 
> Thank you for checking on this Manu!! OK, I spoke with Claudia about this -- because the coverage is so low (like 1/4 of all counties), we think it's probably better to switch back to 5-year estimates for the overall values as well. Is this something you can do?

Absolutely! I'll have this done in the next hour

tinatinc:
> @tinatinc I switched to the 5-year ACS for overall and it greatly improved data completeness, especially for more recent years.

Yep, looks great Manu! I think this is the right approach for now. Just noting one last thing: the counts in the final files are slightly off..
We have 6 years (2016-2022 without 2020), 3143 counties, 486 cities.
So the county subgroup file should have 94,290 observations, but only has 94,280 (10 obs less).
The city subgroup file should have 56,574 observations, but only has 56,569 (5 obs less).

Same with the overall figures -- years 2016-2019 seem to have 3142 counties, and 2021 and 2022 have 3144.
2016 and 2017 has 485 cities, but 2018-2022 have 486. These are probably where the subgroups discrepancies are coming from, too. I don't really see an error or random drops in the code.. do you already know what's going on there?

documentation/overview.md
```diff
@@ -1002,20 +1002,29 @@ Please note that the denominator we use is the living wage for a single full-tim
...
+`digital_access_city_all.csv`, `digital_access_county_subgroup_all.csv`,   `digital_access_city_subgroup_all.csv`,
+* **Data Source(s):** Census Bureau's 5-year ACS.
+* **Notes:** The unit of analysis for the overall and income subgroup is 
+households, whereas the unit of analysis for the race/ethnicity subgroup is
+people in households. 
+* **Data Quality Index:** `1` means the differnence between the estimate and the
+lower bound estimate is less than 0.1, `2` means the difference is between 0.1
+and 0.2, and `3` means the difference is greater than 0.2. 
 * **Limitations:**
 * **Missingness:** There are no missing county or city level observations.
```
tinatinc, line 1019: This is no longer true, right? (re: missing observations), so let's remove or update

08_education/digital_access.qmd
```diff
@@ -0,0 +1,957 @@
...
+  skim
+)
+```
+
+# Create data quality flags
+
+We can use the width of the MOE to determine quality. If the value minus the
+lower bound is \<0.1, then the quality is strong (\_quality=1); if it is =\>.1
+and \<.2, the quality is marginal (\_quality=2), and if it is \>=.2, then the
+quality is weak (\_quality=3). These thresholds apply to both the 1 year and 5
```
tinatinc, line 470: not urgent, but since we are going with 5 year, let's update this language back to just 5 year

tinatinc:
Note for posterity: all inconsistency in observations are due to the crosswalk files, so we are leaving as is for now

PR #299, opened 2024-03-05
Iss201 - Income metric

### LINKED ISSUE #201: 1. Update household income at the 20th, 50th, and 80th percentiles
Pillar: Rewarding Work
Predictor: Opportunities for Income
Metric: Household income at 20th, 50th, and 80th percentiles

- [ ] Rename to `pctl_income_20`, `pctl_income_50`, `pctl_income_80`, and `pctl_income_quality`. 
- [ ] Add confidence interval
- [ ] Add IPUMS API
- [ ] Add 2022 data
- [ ] Make the data quality flag `NA` any time the metric is missing.

jwalsh28:
Closes #201 

2024 Update to the Income metric.

1) Added 2022 1-year data to overall calculation.
2) Added race/ethnicity subgroup.
3) Added confidence intervals (upper and lower bounds)
4) Transition from HMISC package for quantile calculations to the srvyr package. 

Broke the program out into four scripts:
- One for place level data overall
- One for place level data with subgroups
- One for county level data overall
- One for county level data with subgroups 

These programs take a while to run, particularily the subgroups. I hihgly recommend running on the SAS server. There is a lot of duplicate code between the four scrips but due to size they are broken out.


kmartinchek:
Below, please find some notes from my review of the county all file. These also apply to the place all file. I was able to run both of these files without errors on desktop-- but was not able to run them on the servers as you recommended without errors.
1. Can you clarify in line 146 what you mean about the vacancy results?
2. On line 237 you note that you are creating data quality variables-- can you explain what this data quality variable captures (e.g., the quality of crosswalk vs. the data quality of final metrics) to alleviate potential confusion with later quality variables?
3. I see that there are some afact values that are NA-- it could be helpful to explain why this might occur and check where these areas are to assess if any areas are excluded that aren't supposed to be.
4. In reviewing the function for the calc_income_quantiles, it is hard to assess if this is put together properly because there is no explanation of the decisions made in the calculations (to match between what we want to calculate and how we want to do it and how it is done in code).
5. NA data should also be NA/missing in the data quality variables (I think, based on the data quality standards listed on the GitHub site). I would confirm if this is right and make this change if needed.
6. Also related to data quality flags-- the explanation of the missing data flags does not seem to match the function. For example, you just use the geographic_allocation_quality variable to determine data quality when you mention that effective_samples also matter. It is unclear to me how you are account for this and why it isn't included in the function.
7. I see you generated a longitudinal file, but not the recent data file, per the GitHub guidelines. Also, in the output file, you have the FIPS codes for state/county (or place) but do not include the state name or county name, which are part of the final formatting.

kmartinchek:
I now have reviewed the county subgroup file as well. Many of the comments listed above apply to this file, because they have similar code bases. In addition, I have the following questions (which also apply to other scripts):
1. I see in the final output file the final year is 2021. Is that right? It seems like other metrics go up to 2022 data.
2. The final output files are not formatted as FIPS codes (with leading zeros) for state and county.
3. I would recommend adding to all files unit tests to check that each year has the expected number of observations for counties/places. Other unit tests may be helpful, including those that confirm quality variables are NA when the metric is NA, that each county/place has all subgroup rows, etc.
4. Shouldn't there be a quality variable for all metrics? In the subgroup file, I only see one overall.
5. I would add a note at the beginning of the file about how users will not be able to have "umf" in their filepaths at all, per our discussion on issues in generating/replicating the metric. Also a note on file size and need for server/more space for computing, as the subgroup files in particular (although the overall files as well) do not run on desktops of standard computers.

kmartinchek:
No new comments on the place subgroups file, although the comments shared earlier also apply to this script and output.

jwalsh28:
@kmartinchek Thank you for the feedback, I made changes based on your comments and commited the new files!

kmartinchek:
I finished running the subgroup files. At line 615, the ggplot code has a bug, and I get the error-- 
image

I also noted that the subgroup file ends in 2021 while the overall ends in 2022. Is that right?

Otherwise, in checking the file, the same comments as above apply.

kmartinchek:
I have checked all of the code scripts and output at this time and replicated the county overall and subgroup results. I found that these files ran without errors and the output had the expected number of observations and fit the GitHub guidance for creation of these files. As discussed in earlier comments, these files do not have state, county, or place name variables included (which was cleared by Aaron/Claudia) and uses pre-2022 county equivalents for CT, so wanted to note those were as expected. In my review, I wanted to note:
1. I saw that the place and county subgroup files include 5 year samples for the 2018 and 2021 ACS. I noted this in earlier comments, on 3/6 and 3/15, but wanted to confirm these should not have 2022 data. It looks like IPUMS released their 2022 5 year samples on 2/6/24, but unsure if this was excluded for other reasons.
2. In my earlier comment on 3/6, I asked for more detail on the decisions made in the calc_income_quantiles function to check how the code aligns with the intended calculation. It might be helpful to clarify that the as_survey_rep function argument choices are from IPUMS guidance on using replicate weights, accessible here: https://usa.ipums.org/usa/repwt.shtml (I looked this up to confirm why you chose particular specifications in the arguments). But I would like to confirm the choice for argument qrule = "hf3" in the survey_quantile function. 
3. I would like to confirm the construction of the new data quality flags to make sure I understand what you are doing here. I see that the geographic_allocation_quality variable is from the crosswalk and ranges from 1-3. The text of the the data quality flag notes that it relies on this variable and the effective sample. Am I right in that the flag is NA when the effective sample is below 30 (the metric is missing), but otherwise takes on the value of geographic_allocation_quality? Unsure if effective samples otherwise impact the value of the data quality flag, since that wasn't entirely clear to me.

kmartinchek:
John + I met briefly and confirmed that:
1. 2022 was not planned to be included in the subgroup data because 4/5 would be similar to the prior 2021 release.
2. The argument qrule = "hf3" in the survey_quantile function is needed to replicate the treatment of weighted quantiles executed in prior data updates. This will be clarified in documentation but the code is appropriate as-is.
3. The data quality variable construction logic is as I described above and as such, aligns with the code.

All in all, it looks as expected and is replicable. I also was able to run and replicate the place overall and subgroup files without errors in the interim time.

kmartinchek (approved):
See comment history on PR. All is as expected as noted in comments, so I approve.

PR #302, opened 2024-03-14
Iss220 -- broadband access

malcalakovalski:
Reformat final files to match README standards and relabel subgroups. In particular,

- Drop 2016 since it is all missing values and update documentation on years accordingly
- Remove the percentage sign in output share values
- Remove state_name & county_name or place_name from the final files
- Change the race-ethnicity labels from "White, Non-Hispanic" and "Black, Non-Hispanic" to "White" and "Black"
- Make sure subgroup_type comes before subgroup in the ordering of the variables in the output files
- Change the income subgroup labels to: Less than $50,000, and $50,000 or More to clarify where $50k falls into


malcalakovalski:
- Drop 2016 since it is all missing values and update documentation on years accordingly
- Remove the percentage sign in output share values
- Remove state_name & county_name or place_name from the final files
- Change the race-ethnicity labels from "White, Non-Hispanic" and "Black, Non-Hispanic" to "White" and "Black"
- Make sure subgroup_type comes before subgroup in the ordering of the variables in the output files
- Change the income subgroup labels to: Less than $50,000, and $50,000 or More to clarify where $50k falls into

tinatinc:
All looks good to me!

PR #310, opened 2024-03-18
Iss202 - Housing Affordability

rogin123:
create new branch for Rob to review


02_housing/1a_housing_place.R
```diff
@@ -294,7 +293,7 @@ households_2022 <- left_join(microdata_housing, place_income_limits_2022, by=c("
 # for renter and owner subgroups, this is just the population of renters or owners at each income level
 
 # create new variable 'Affordable80AMI' and 'Below80AMI' for HH below 80% of area median income (L80_4 and OWNERSHP)
-# if OWNERSHP is not equal to 1 or 2, leave as NA
+# if OWNERSHP is not equal to 1 or 2, leave as NA
```
rpitingolo, line 296: We have an issue on the case_when code:

```
Affordable80AMI_all =
           case_when(# deal with cases when RENTGRS and OWNCOST are 0
             RENTGRS == 0 ~ 0,
             OWNERSHP==2 & ((RENTGRS*12)<=(l80_4*0.30)) ~ 1,
             OWNERSHP==2 & ((RENTGRS*12)>(l80_4*0.30)) ~ 0,
             OWNERSHP==1 & ((OWNCOST*12)<=(l80_4*0.30)) ~ 1,
             OWNERSHP==1 & ((OWNCOST*12)>(l80_4*0.30)) ~ 0)
```

I believe that if the record is for an owner (OWNERSHP==1) that it is a zero on the variable RENTGRS. So essentially you are missing all ownership cases because you are making them zero off the bat. This also applies to the 50 and 30 AMI code sections.

It looks like the county code is correct. However I feel like a better and safer way to code this would be to calculate Affordable80AMI_renter and Affordable80AMI_owner first, then calculate Affordable80AMI_all with the formula: if Affordable80AMI_renter==1 OR Affordable80AMI_owner==1. 

This needs to be fixed here as it is used in the 2a and 2b scripts.

rpitingolo (changes requested):

02_housing/1a_housing_place.R
rpitingolo: However you decide to fix it, I feel like we should make sure it is coded the same way on the county script for consistency.

awunderground: IPUMS loads data labels, which can be converted and used directly. FWIW, I think this limits issues with mixing up 0/1. It may not have exactly solved this problem, but it may have been clearer.

rpitingolo: OK I re-checked all of the variables to make sure the _renter and _owner version are the same as the _all and it check out. Going to resolve this now.

02_housing/1a_housing_place.R
```diff
@@ -58,7 +57,7 @@ library(tidylog)
 
 # Either run "0_housing_microdata.qmd" OR: Import the already prepared microdata file 
 # this one should already match the PUMAs to places
-acs2022 <- read_csv("data/temp/2022microdata.csv") 
+acs2022 <- read_csv(here::here("data","temp","2022microdata.csv"))
```
rpitingolo, line 60: It looks like you updated the read_csv to use `here::here` but isn't the data in a temp folder in the root directory not the housing subdirectory? I had to edit this to get it to work. It also doesn't look like you changed this on the 2a and 2b programs so whatever we decide should make it consistent.

02_housing/2a_affordable_available_place.R
```diff
@@ -110,7 +110,7 @@ available_2022 <- available_2022 %>%
 # (4) calculate the number of affordable and available units at each income level per 100 households
 # this is the number of units that are rented to people at or below each income threshold
 # plus the number of vacant units at that income threshold over the total number
-# of households in each income bracket
```
rpitingolo, line 113: Now that we resolved the issue on the 1a script this calculation now seems correct. However,
1. My understanding (from Claudia) is that these variables should be named share_ rather than rate_ to be consistent with the other final files
2. They should not be multiplied by 100, again to be consistent with the other final files

02_housing/2a_affordable_available_place.R
```diff
@@ -161,7 +161,7 @@ available_2022_subgroup <- available_2022_final %>%
   # remove leading underscore and capitalize words
   mutate(subgroup = str_remove(subgroup, "_") %>% str_to_title(),
          subgroup_type = "tenure") %>% 
-  # join with affordable values for final file
```
rpitingolo, line 164: My understanding (from Claudia) is that these final files should match the format of the previous final files. I believe this means that what is currently rate_ should be switched to named share_ and these affordable values do not need to be merged on since they already exist in the other files.

rpitingolo (changes requested):
@rogin123 some things to look at in the 2a script. Thanks!

02_housing/2a_affordable_available_place.qmd
```diff
@@ -0,0 +1,251 @@
...
+  mutate(across(matches("share_.*quality$"),
+                list(share_affordable_available = ~.), 
+                .names = "share_affordable_available_{str_remove(.col, 'share_affordable_')}"), 
+         # subpress counties with too small of sample size
+         across(matches("share_affordable_available.*ami$"), \(x) if_else(is.na(get(cur_column() %>% paste0("_quality"))), NA, x))) %>% 
+  # arrange columns and drop unneeded ones
+  select(year, state, place, subgroup_type, subgroup, matches("available_30"),  matches("available_50"), matches("available_80"))
+
+# export our file as a .csv
+write_csv(available_2022_subgroup, here::here("02_housing/data/available_2022_city.csv"))
```
rpitingolo, line 180: Is it right that the output should have 1,458 rows? 3 for each of 486 places?

02_housing/2a_affordable_available_place.qmd
```diff
@@ -0,0 +1,251 @@
...
+write_csv(available_2022_subgroup, here::here("02_housing/data/available_2022_city.csv"))  
+```
+
+## (5b) overall file
+
+```{r}
+# combine with share_affordable file 
+affordable_2022_overall <- read_csv(here::here("02_housing/data/housing_2022_city.csv"))
+
+# keep what we need
```
rpitingolo, line 189: I don't understand the purpose of this chunk. Since we already have the affordable metrics output in another file, I don't know what the purpose of merging them back on here is? We also now have 3 rows per place even though they are all supposed to be in the subgroup "all"?

rpitingolo (changes requested):

02_housing/2a_affordable_available_place.qmd
```diff
@@ -0,0 +1,251 @@
...
+  pivot_wider(
+    names_from = available, 
+    values_from = value
+  ) %>% 
+  # clean subgroup names and add subgroup type column 
+  # remove leading underscore and capitalize words
+  mutate(subgroup = str_remove(subgroup, "_") %>% str_to_title(),
+         subgroup_type = "tenure") %>% 
+  # join with affordable values for final file
+  left_join(affordable_2022_subgroup) %>%
```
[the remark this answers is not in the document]
rogin123: This was a great catch! I merge in the `affordable_2022_subgroup` file to get the data quality variable which is the same for both the `share_affordable_` and `share_affordable_available_` values since they come from the number of households below each AMI. So this section merges in the quality flags and then renames them to make sense for this variable. I accidentally dropped the subgroup column before I merged though so it was joining wrong but fixed that.

02_housing/2a_affordable_available_place.qmd
rogin123: Yes, but just noticed that I messed up the naming when I saved out the files and switched which one I called subgroup and overall

02_housing/2a_affordable_available_place.qmd
rogin123: Same as above, using to add the data quality flag.

02_housing/1a_housing_place.R
rpitingolo: I am still having some issues with this `here::here` code. It keeps trying to use a subfolder where the data doesn't exist.

02_housing/2a_affordable_available_place.qmd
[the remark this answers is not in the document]
rpitingolo: OK. I didn't realize that was the intention of the join. I think it's OK now.

02_housing/2a_affordable_available_place.qmd
rpitingolo: OK I think now that the previous issue is resolved this one is too.

02_housing/2a_affordable_available_place.qmd
```diff
@@ -0,0 +1,253 @@
...
+                list(share_affordable_available = ~.), 
+                .names = "share_affordable_available_{str_remove(.col, 'share_affordable_')}"), 
+         # subpress counties with too small of sample size
+         across(matches("share_affordable_available.*ami$"), \(x) if_else(is.na(get(cur_column() %>% paste0("_quality"))), NA, x))) %>% 
+    # arrange columns and drop unneeded ones
+  select(year, state, place, matches("available_30"),  matches("available_50"), matches("available_80"))
+
+
+# export our file as a .csv
+write_csv(available_2022_overall, here::here("02_housing/data/available_2022_city.csv"))
```
rpitingolo, line 208: The final data is available_2022_overall, correct? My understanding is that it should have 6 variables (3 metrics plus 3 quality flags). This has 3 metrics and 6 quality flags. Perhaps @cdsolari can weigh in as I'm not 100% certain about this.

rpitingolo (changes requested):

02_housing/2a_affordable_available_place.qmd
cdsolari: @rogin123 @rpitingolo it looks like the available_2022_city file for the overall values is borrowing the format of the subgroup file. It should only be one census place per row, and it shouldn't have the subgroup_type or subgroup variables in that one. It should look more like your county version of that file (though see below because there are issues there too). I think this will be addressed it you limit it to just the All rows, but note that you have some strange things going on with your all rows because you have 3 per place, and not all of them have the same values. If you have them on multiple rows, I'd expect them to be duplicates and you can just remove the renter or owner subgroups and deduplicate on the all and then remove the two variables to indicate subgroup. But, I'm not sure what value is the correct "all" value because some are identical while others are different. The main difference seems to be some where you have more values that are NA. Let me know if I can help iron this out. I also see a similar problem in the county file. It looks like you kept the "all" rows and removed the subgroup_type and subgroup variables, but you still have 3 rows per county, and they aren't showing the same for each. I have a feeling you have code that maybe progressively gets more restrictive, but instead of applying it to the same row, it creates more rows that is each progressively more restrictive. I'm happy to show  you what I'm seeing.

rpitingolo: @cdsolari regarding the 3 rows per place issue. How are you looking at the data? Are you looking at the CSVs pushed to github? Or looking at the dataframe in R Studio? This was an issue I previously flagged for Amy that I think is now resolved so want to make sure we are all looking at the same thing.

rpitingolo:
OK so we have a problem where the metric for affordability at 80 AMI is lower than at 50 AMI in a substantial number of counties. I ran a check with this code:

```
check <- available_2022_overall %>%
  mutate(flag_30_gt_50 = if_else(share_affordable_available_30_ami > share_affordable_available_50_ami, 1, 0),
         flag_50_gt_80 = if_else(share_affordable_available_50_ami > share_affordable_available_0_ami, 1, 0))
```
         
 And then number of flag_50_gt_80 make up the majority of cases, so that doesn't pass the smell test. 

I tried to trace this back through the code and couldn't the issue but when I checked the data itself I think I found it there. I can't run the code straight through because it breaks on this line:

`households_2022 <- read_csv(here::here("02_housing/data/temp/households_2022_county.csv"))`

I am able to run it using the households_2022_county.csv from the root folder (not the 02_housing subfolder) but I think that is an outdated file and the reason why we eventually wind up with the issue I originally started with. 

Since these scripts rely heavily on reading and writing CSVs we need to make sure all the CSVs are in the correct place and any old or outdated CSVs are deleted. I should be able to press "run all chunks" and it runs through without breaking. 

@rogin123 can you please make sure the correct CSVs are in the correct places, delete any outdated ones, and make sure all committs are pushed? Can you also run that QC code I pasted above on both the place and county data to make sure it looks OK? There are a small number of cases on the place data where it's flagging but we can check that separately as it might be legitimate. 

@cdsolari once this is resolved I can re-check the metrics.

Thanks!

02_housing/2a_affordable_available_place.qmd
cdsolari: @rpitingolo You're right, I was looking at an older version of the file. Thanks for catching that!

rpitingolo:
I've spot checked both place and county CSVs and don't see any more major issues, so I am going to approve this now.

rpitingolo (approved):

PR #311, opened 2024-03-18
Iss224 -- Home Value (Fix output files)

malcalakovalski:
- Adjusts values to take the form of 2 values between 0-1 with a leading zero separated by a colon (Instead of 2 values between 0-100 separated by a colon).
- drops `state_name` and `county_name`/`place_name`
- changes age labels to:  "Under Age 45"; "Age 45 and Over"


cdsolari (approved):
I reviewed the final files to make sure that we are moving from showing just the tenths place to showing to the thousandths place. The files look good!

PR #317, opened 2024-03-21
Iss223 - injury death rate fixes 

vpancini:
- Drop these 4 variables: `state_name`, `county_name`,  `rate_injury_death_se`, `rate_injury_death_crude`
- Change the `_lb` and `_ub` values to `NA` for instances where `rate_injury_death` is `NA`. These are instances where the `rate` was originally coded as `Unreliable` by CDC WONDER because the death count was less than 20. We changed the `rate` to `NA` but didn't adjust the bounds, which had non-`NA` values.


cdsolari (approved):
This was just to drop some excess variables and sort them. It looks good!

PR #318, opened 2024-03-25
Iss204 - population to pcp ratio fixes

vpancini:
- Dropped the following variables: `state_name`, `county_name`, `ratio_population_pc_physician_flag_low_pop`

- In 2019, there were some cases where we had a value for this metric, but the data quality is `NA`. These were almost entirely (all but 2 cases) where the `ratio` was 0 but population was greater than 2,000, so we were assuming these were “false zeros” and suppressing the values (the `ratio` values themselves should have also been `NA` in these cases in 2019). In the initial update of this metric we did not apply this suppression rule, but in this push I did apply it. So, now, if the population is less than 2000 and 0 PCPs are reported, we assign a ratio value of 0. If the population is greater than 2000 and 0 PCPs are reported, we assign a ratio value of `NA` and assume that these are "false zeros." For the later cases, the `quality` value is also `NA`. For the former cases, the `sensitivity` value that we calculate and use to construct `quality` is `Infinity` because of a divide by zero error. I recode these values to equal the maximum defined value of `sensitivity` which leads all of these observations to receive a low `quality` value. Note that in 2019, all of these observations (the assumed “true zeros” with 0 PCPs and pop < 2000) had high `quality` values. I looked at the 2019 code and could not figure out why, but I think it was an error, and may be that the way R deals with infinity values has changed since the last update. I explained this in my code as well.

- There were two cases where the `ratio` value in 2019 was `1589:1` and the quality value is `NA` - these were counties (Jasper and Newton Counties, Missouri) where CHR did not have a count of primary care physicians for, but did have a ratio. Using the raw AHRF data, we have ratio values and quality values for both of these counties in all years.


cdsolari (approved):
All of the issues we noted were taken care of. This checks out.

PR #323, opened 2024-04-10
Iss304

awunderground:
Is this going to break the metrics?


awunderground:
I reworked the place file to use the same PEP data that you used. I think the populations are mostly the same but I correctly sorted the file. Please review with a discerning eye!

PR #327, opened 2024-04-12
Iss324

awunderground:


10_construct-database/11_construct_county_all.qmd
```diff
@@ -0,0 +1,546 @@
...
+
+### Full Database
+
+Evaluate the quality variables. 
+
+```{r}
+db |>
+  select(ends_with("_quality")) |>
+  assert(in_set(1, 2, 3), everything())
+
```
jwalsh28, line 219: What are we supposed to be taking away here?

10_construct-database/11_construct_county_all.qmd
```diff
@@ -0,0 +1,546 @@
...
+    pctl_income_80
+  ) |>
+  assert(
+    within_bounds(0, Inf), 
+    share_debt_coll, 
+    share_debt_coll_ub, 
+    share_debt_coll_lb
+  ) |>
+  quiet()
+
```
jwalsh28, line 239: The pctl_income variables have CIs - why are we not using the test bounds function on these variables? Im guessing it's becuase for these vlaues there is the possibility of a CI being Inf even when the variable is included and it would throw off the test.

geographic-crosswalks/create-county-populations.qmd
```diff
@@ -0,0 +1,256 @@
...
+  geom_point(alpha = 0.1) +
+  facet_wrap(~ year) +
+  labs(title = "Population change in different years")
+
+population_test |>
+  ggplot(aes(population, prop_pop_change)) +
+  geom_point(alpha = 0.1) +
+  facet_wrap(~ year) +
+  labs(title = "Proportional population change in different years")
+
```
jwalsh28, line 194: These visuals are really helpful, I love them - well done

jwalsh28 (changes requested):
The population files look great to me, very helpful that this is so well documented now. 

Note the missing files from the construct_county_all file. Once these are added this should be good to go!

10_construct-database/11_construct_county_all.qmd
awunderground: All quality variables should only have the value 1, 2, or 3. This will throw an error if a quality value is 4 or 2.2.

10_construct-database/11_construct_county_all.qmd
awunderground: This was an oversight. Good catch!

PR #335, opened 2024-04-18
Iss221

rogin123:
remove duplicates and change the zcta to place calculation to sum rather than weighted average


malcalakovalski:
This looks great to me! I confirmed that the duplicates were removed and reviewed that the code reflects the corrected logic to calculate the numerator

malcalakovalski (approved):
This looks great to me! I confirmed that the duplicates were removed and reviewed that the code reflects the corrected logic to calculate the numerator

PR #343, opened 2024-04-24
Fix naming issues with combine files

jwalsh28:
-	Rate arrests race groups (should not include white, non-Hispanic) 
-  Communities – maybe try “Neighborhoods”


10_construct-database/61_construct_place_all.qmd
cdsolari, line 1: Can you change the "asian_other" to share_desc_rep_other, share_desc_rep_other_quality in all your files? This is what we told Graphicacy to expect. If you feel strongly about it, we can communicate this change and it shouldn't be too late!

cdsolari (changes requested):
We still have places in several files where the "asian_other" isn't changed to share_desc_rep_other, share_desc_rep_other_quality. Is that something we can change? I also know that Aaron made a last minute decision to change all the racial sub-area labels to something like white communities, communities of color, and mixed communities. Can we chat about this? It is lowest priority though!

cdsolari (approved):
Looks like all the intended changes worked!

jwalsh28:
Great! Leaving this PR open for now just in case we stumble across anyhting else as we are finalizing tests.

PR #347, opened 2024-04-25
Iss346 - Updating transit trips weighting

tinatinc:
Reverting transit trips collapse to a mean instead of a sum. Affects county subgroup, city, and city subgroup files. And their outputs, which I have not run


06_neighborhoods/Transportation/final/transit_trips_all_subgroups_county.csv
cdsolari, line 1: This file has cases where the state is NA and county is NA, with NA values that needs to be cut from both 2015 and 2019.

cdsolari (changes requested):
In looking at the transit_trips_all_Subgroups_county.csv, there's an extra row with NA for state and county for both 2015 and 2019 that needs removing. Thanks!

PR #352, opened 2024-04-30
Check median quality

judah-axelrod:
@jwalsh28  - `medians-long-county.qmd` is the relevant file to review. I just added a statement to print out the suppressed county-level medians at the end. Currently assuming >15% population missing or poor quality is the threshold for suppression but @awunderground let me know if you'd like to see the more conservative 10% version.

Let me know if you have any questions!


jwalsh28 (approved):
This looks good to me. One structural comment, it would be helpful if the indivdiual calculations were broken out into seperate chunks in the QMD document with explenations around what each one does. Would have helped me wrap my head around each step faster.

judah-axelrod:
> This looks good to me. One structural comment, it would be helpful if the indivdiual calculations were broken out into seperate chunks in the QMD document with explenations around what each one does. Would have helped me wrap my head around each step faster.

I will make this change, good call!

PR #353, opened 2024-05-17
Iss209 - divide by 100

rpitingolo:
@cdsolari updated code and CSVs on this branch where the cost metric is divided by 100 and now has a value between 0 and 1.


cdsolari (approved):
The overall and the subgroup all match, and the figures are divided by 100 to be between 0-1.

PR #358, opened 2024-07-01
Fix small weighting issue on the rent burden varaible and correct joi…

jwalsh28:
…n error that was resulting in NAs for the county file


cdsolari (approved):
Reviewed the changes. See good catches for weighted means and merging on the right county codes. These changes resulted in matching values for DC city & county, and all the other jurisdictions we check that are both a city and a county and should have identical numbers (and they do!). They do not line up exactly with the NLIHC gap report figures, but they shouldn't because their geography is Metro Area/MSA, but we did a cross-check just in case to see they they aren't wildly off, and they are the best version we know. We are still waiting on a response from NLIHC to share their methodology on the gap report. It seems like they may not use HUD-published FMR data in their calculations like we do, but it isn't clear. But, this is accurate for what we are trying to calculate for the geographies we are calculating.

PR #372, opened 2024-08-16
Add count_quality flag to homelessness metric

### LINKED ISSUE #369: Create a count and share quality flag for homelessness
Bring back in updated code currently living on Main (metric lead pushed to wrong branch)

jwalsh28:
Closes #369  I just added the variable in the combine file - I actually had a tough time finding the program that creates this data.. in the future we should add it directly there.


cdsolari (approved):
I checked the code and that in the final file for county longitudinal and downloaded the csv and confirmed that in all cases count_homeless_quality=share_homeless_quality.

Issue #374, opened 2024-08-19
Set up geographic harmonization for next update


awunderground:
We need to deal with at least two types of geographic harmonization:
1. How geographies relate from year-to-year. For example, the Census Bureau [switched](https://www.ctdata.org/blog/geographic-resources-for-connecticuts-new-county-equivalent-geography) from reporting data for eight counties in Connecticut to nine planning regions in Connecticut for 2022 data.
2. How geographies where data are reported (e.g. PUMAs and ZCTAs) relate to to our target geographies (counties and populous census places).

We need to develop a comprehensive plan for addressing both of these questions and create “target” files and crosswalk files that all developers will use to consistently create metrics over times.

**First**, we hope to create a Quarto document that articulates a plan to answer big questions:

- What are all of the geographic crosswalks we need?
- What time perspective should we use? For example, should we report Connecticut data for the nine planning regions before 2022 or use the eight counties?
- We use a population threshold to identify large cities. Should this change over time as cities grow and shrink?

This document should include the number of geographies in each year, justifications for decisions, and documentations of important changes to geographies.

**Second**, we need to update the repo to contain standardized target files and crosswalks that are described in the step 1.

malcalakovalski:
I noticed a couple of files in the `geographic-crosswalks/data` folder are unused. For example, none of the files starting with `long-xwalk` are used by any script. It looks like these were [added](https://github.com/UI-Research/mobility-from-poverty/commit/e9e4ed459bb0eaac2cc0be6654c6d0df6ba3bc92) in April 2023 but with no accompanying script. Likewise, the `names-to-fips` files aren't called in any scripts.

Should these be deleted from the repo to avoid confusion?

malcalakovalski:
I wonder if it would be beneficial to standardize the name of the crosswalk files to be something like `{source-geography}-{target-geography}-crosswalk_{year}`.

To prevent breaking existing code that relies on crosswalks with old file names, we could simply start by encouraging adopting the new file name convention in the documentation for now and edit the scripts at a later date (maybe next year after the metrics are updated?)

malcalakovalski:
@awunderground I also noticed that the process of downloading the tract to place 2020 crosswalk from geocorr is documented in `poverty-exposure-2021.qmd` and in `race-ethnicity-exposure-city-2021.qmd`. This strikes me as duplicative and hard to find!

[...]

PR #384, opened 2024-09-16
Set up geographic harmonization for next update

malcalakovalski:
This PR is a work in progress but currently includes: 

- A quarto document with a draft explaining year-to-year changes in the number of counties and places
- An update to `create-place-populations.qmd` so that it pulls now available 2023 population data instead of using 2022 data for 2023. I did this while investigating the population thresholds for places, but let me know if this belongs in a different PR. **Note:** There are some linting differences in this commit which can be disregarded.

It's currently missing a description of all the geographic crosswalks we need, which I will tackle this week!


malcalakovalski:
I also have thought and read more on the questions listed in #374. In particular:

-   **What time perspective should we use?** I believe we should report Connecticut data for the nine planning regions before 2022 to be compatible with future census data releases. Is there any concern with this approach?
-   **We use a population threshold to identify large cities. Should this change over time as cities grow and shrink?** While I'm inclined to say yes, I'm unsure of how to operationalize this in a reasonable way. The only relevant guidance I found from the Census is in their [Urban Area classification criteria](https://www2.census.gov/geo/pdfs/reference/ua/Census_UA_CritDiff_2010_2020.pdf) which says that the 2010 Census used a 50,000 threshold for Urban Areas while the distinction between Urban Areas and Urban Clusters was removed in the 2020 Census. Therefore, I believe we should keep the current threshold for now unless we find a sensible way of varying it year-to-year.

malcalakovalski:
I noticed a couple of files in the geographic-crosswalks/data folder are unused. For example, none of the files starting with `long-xwalk` are used by any script. It looks like these were [added](https://github.com/UI-Research/mobility-from-poverty/commit/e9e4ed459bb0eaac2cc0be6654c6d0df6ba3bc92) in April 2023 but with no accompanying script. Likewise, the `names-to-fips` files aren't called in any scripts.

Should these be deleted from the repo to avoid confusion?

malcalakovalski:
I wonder if it would be beneficial to standardize the name of the crosswalk files to be something like `{source-geography}-{target-geography}-crosswalk_{year}`. 
 
To prevent breaking existing code that relies on crosswalks with old file names, we could simply start by encouraging adopting the new file name convention in the documentation for now and edit the scripts at a later date (maybe next year after the metrics are updated?)

geographic-crosswalks/geographic-harmonization-guide.qmd
```diff
@@ -0,0 +1,90 @@
...
+### How many places are in our data?
+
+Our final data contains incorporated places with populations greater than 75,000. The number of places in our data is 485 for 2015-2017 and 486 from 2018-2022. This is because the Census Bureau [added South Fulton City, GA in 2017](https://www.census.gov/programs-surveys/acs/technical-documentation/table-and-geography-changes/2018/geography-changes.html).
+
+```{r}
+places_data <- read_csv(here::here("data/61_mobility-metrics_place_longitudinal.csv"))
+
+places_data |>
+  count(year)
+```
```
jwalsh28, line 83: Do we have the same 485/486 places each year? Are any dropping out and in?

geographic-crosswalks/geographic-harmonization-guide.qmd
```diff
@@ -0,0 +1,90 @@
...
+
+Connecticut’s historical eight counties stopped serving as functioning governmental or administrative units in 1960. Currently, Connecticut has nine [Regional Councils of Government](https://portal.ct.gov/OPM/IGPP/ORG/Planning-Regions/Planning-Regions---Overview) (COG) that carry out regional planning and service delivery activities similar to those performed by county-level governments in other states.Planning regions are administrative entities that have have the same boundaries as the state’s COGs.
+
+The Census Bureau used to release data based on Connecticut’s eight counties, but starting in 2023, it switched to reporting data for the nine planning regions, following a 2019 request from the state’s Office of Planning and Management. Since new data products will include planning regions, *we should make sure to use them instead of the historical counties moving forward*. 
+
+See the map below for the planning region boundaries, the towns within each planning region, and historical county boundaries (delineated by thick white borders)
+
```
jwalsh28, line 66: Based on the comparison between prior counties and these planning regions, do you suggest we make a clean break and start just showing the planning regions OR should we try to crosswalk backwards to the counties so we can have a coherent time series for these?

jwalsh28 (changes requested):
Hey Manu, left a few thoughts - I think this is a good start. Can you render the QMD document so it can be viewed as an HTML document?

Also, can you include some analysis of our current crosswalk file? I would be interested if it aligns with the geographies in the final combined data sets - I think some metrics used their own crosswalks so there may be some disconnect.

geographic-crosswalks/geographic-harmonization-guide.qmd
malcalakovalski: I'm going back and forth a bit on whether it would be best to crosswalk past years of data into planning regions, or make a clean break. I'm slightly leaning towards the first option since counties haven't been functional governmental units in CT since 1960. Do you know if there are any data quality concerns related to revising past data to make a consistent time-series? Would this cause problems for users of our data who want to merge it with their own data, or do you think adding an explanation about this change for users would suffice?

jwalsh28 (approved):
Looks great Manu!

PR #385, opened 2024-09-17
First stab at the final data evaluation function. This includes a tem…

jwalsh28:
…plate for the final data expectations form which is used in the test function. It also includes an example in QMD format based on the housing affordability final data.

Note that there is also a page in the Wiki now describing this function and how to apply it. Some outstanding questions below:

For final data expectations - do we have them read in the form at top of each metric program or is it fine as an argument in the function?
Where should the final data expectations forms live? Right now I'm encouraging the final data folder for the relevant metric.
And where should the final data expectations template live? I have it in the functions/testing folder as of now.
Where should the test function example live? I have it in the functions/testing folder right now.


wcurrangroome (changes requested):
See overarching feedback in a comment on the `.qmd`. Looks good, JP!

wcurrangroome (approved):

PR #389, opened 2024-09-19
Remove zombie columns from combined data sets 61 and 63

### LINKED ISSUE #388: Remove zombie columnds from combined files
There are two variables that made it into the combined files which do not belong - these were mislabeled in the final files and snuck through tests. Fixing these by removing them in the combined codes. See slack below identifying the issue: 

Looks like in file 63 there is a city column that I missed, the data in it is accurate but clearly it should be not be there. It snuck in via the debt in collections 2021 file.
For state_place, similar thing this is sneaking in with the electoral turnout data. This is a great catch, definitively raises some thoughts for the tests Aaron and I run on the final combined files.

jwalsh28:
Closes #388 

I removed these columns from the raw final data sets in the combined process. These should be taken out of the raw data via the program in the next update round.


awunderground (approved):
Looks good!

jwalsh28:
Closes #389

PR #396, opened 2024-10-15
Update git issue template to create final version. Make crosswalk bet…

jwalsh28:
…ter reflect Wiki and change language


cdsolari (approved):
Looks great!

PR #424, opened 2024-10-28
Iss397

wcurrangroome:
This is an interim PR / code review request. There is no need to check the line-by-line code logic yet. Will is going to review the code in comparison to the prior year's code concurrent with this review and make other (small) adjustments as needed. Line-by-line review will be more helpful following this initial review.

**Background:**
- The prior version of the code for this metric was hard to follow, redundant at points, and relied on writing out and reading in datasets repeatedly, even though the computational time for calculating these metrics is minimal. The goals for this metric's update -- to add CIs (potentially) and to back-calculate for a number of years, as well as to update with 2023 data -- made the existing code challenging to work with.

**Updates:**
- Code has been reorganized into two scripts: one that generates all microdata and one that produces all metrics.
- Additional text has been added to clarify what metrics are being calculated, how they relate to one another, and the assumptions used throughout to calculate each metric.
- A number of additional quality checks--some testable assertions, others visualizations and descriptive statistics--have been added.

**Code Review Priorities:**
- [ ] Does the new organization of the code make sense? Are there opportunities to strengthen the documentation or further reorganize?
- [ ] Does the crosswalking process--taking AMI limits and translating them to our geographies of interest (place/county)--make sense?
- [ ] Do the quality checks make sense? Are there opportunities to add more testable assertions?


rpitingolo:
Thanks @cdsolari @wcurrangroome! I'll take a look at this soon.

wcurrangroome:
@rpitingolo -- flagging that in the calculation of affordable renter households in the prior code, there are two things that I have changed:
- For the calculation of `Affordable80AMI_renter` -- the `case_when()` logic was ordered slightly differently than it was for other parallel variables (`Affordable50AMI_renter` and `Affordable30AMI_renter`), leading to inconsistencies in how these variables were operationalized.
- For this series of variables (in the original code), when `RENTGRS == 0`, a value of 0 is returned. I think this is incorrect. When a household's rent is 0, the household should by definition have affordable housing, right?

rpitingolo:
Hey @wcurrangroome Great job. This is extremely high quality code. 

First to address your initial questions:

>  Does the new organization of the code make sense?

 
 Generally yes. However it's not clear to me why we need a separate 0_ script in this new organization? The 0_ simply compiles a function, right? In which case I don't see why it can't either be moved into the 1_ code or if there's a compelling reason to keep it separate, moved into a "supplemental functions" folder or something like that. My issue is that using a numbered scheme implies the code has to be run in sequential order and that each script produces some kind of output. 
 

> Does the crosswalking process--taking AMI limits and translating them to our geographies of interest (place/county)--make sense?

If I understand this correctly you are first creating the income limits for county and only if place is specified in the housekeeping section do you apply the weighting for place. For place it looks like you are doing a weighted mean by population which I think is right. I would want to double check a few cases to confirm it passes a gut check (I haven't done this yet). 

> Do the quality checks make sense? Are there opportunities to add more testable assertions?

Overall I think yes, these checks make sense. However interpreting whether a check is passed or failed is a little less clear. For example for the check `The counts of observations align... `(line 834) I am getting an error on the `stopifnot`. Does that mean the check is failing? Or that the code is incorrect and thus the error? 

I am personally not a fan of using plots as checks. For example the last check in the code uses six plots and eyeballing them they look fairly similar but it's really hard for me to eyeball them and make a confident conclusion that it passes the check.

Have you ever used the package `assertr` ? I just learned about it this week but the person who told me about it seems to think it adds value to their code and data checks. It might be good to use here. 



I have a few other comments that I will type up in a separate comment in a bit.

rpitingolo:
A few more comments @wcurrangroome - 

Overall I like the step by step descriptions you wrote under "Analysis process". Can we use that same numbering / title scheme for the QMD outline? I think it would help with refencing specific sections quickly. 

I feel like there is a lot of important stuff in the "Housekeeping" section and calling it "Housekeeping" might be doing a disservice. For example this is where you would specify county vs. place but it seems kind of buried here. Maybe we can put this parameter in the header? Or a separate section that better calls out that this is where you need to change the value to toggle between county and place.

Similarly this is where the `get_housing_microdata` function lives. To me that's not so much "Housekeeping" as it is something like "Data download" (which is distinct from Microdata import which you do have in a separate section. 

From what I can tell `get_housing_microdata` only runs if the 2022 csv is in the temp folder. We are assuming that if they have that file that they have them all? This is probably fine but I could potentially see an issue down the road if you need to add a new year of data it might not correctly download so long as the 2022 file exists. 

> For this series of variables (in the original code), when RENTGRS == 0, a value of 0 is returned. I think this is incorrect. When a household's rent is 0, the household should by definition have affordable housing, right?

This is right IF a household has no rent. But IIRC a value of zero means N/A not no-rent. However that's only for the variable RENT and I'm not sure if that also applies to RENTGRS. The ipums documentation is not too clear to me on this question. 

> another potential issue: in 0_get_housing_microdata.R, distinct() is applied to microdata.


The microdata should not have duplicates so I think you are right to remove it.

wcurrangroome:
@rpitingolo -- back to you for another round of review. my hope is that the code logic is essentially final at this point, so in contrast to the last round, please review thoroughly as possible. then we can turn to other years' data, CIs, etc.

notes:
- some of the tests error out because there are minor unresolved issues (e.g., South Fulton throws off place counts; the re-produced data are known to be different from the number produced last cycle);
- i've added a rendering function so that we can iteratively produce rendered reports for place/counties, and for different years, but i'm getting some tedious error when running `quarto_render()`. i'll work on this, but for the time being am supplying the params directly in the yaml of the .qmd. (also can't render because of the known errors generated during the QC tests).

rpitingolo:
@wcurrangroome Great job. Many of the code upgrades look good. I did not test `quarto_render()` so this review is based on running code chunk by chunk. 

I think putting the changeable parameters into the header is the right approach, but I also think we need some sort of validation that the values put there are correct. For example I believe the only two allowable values for geography are `county` and `place` but I can put `city` as the geography parameter and the code will mostly run without error but the results are not correct. I think it should be fairly easy to add some kind of check at the beginning of the code. 

For the QC checks, I like the use of the **testthat** package. It seems to be a good way to do straightforward checks. I'm still not entirely certain what I'm supposed to be looking for on the checks that create visual plots. Maybe this is OK if you can look at them and say it's OK but for someone reviewing without as much knowledge of the topic it's tricky. 

I saw the failed check on line 1025 caused by South Fulton. 

I also got a failed check on line 1096 when running for place. The plot on line 1203 also failed to render when running for place. It looks like it has county hard coded into it.

jwalsh28:
> o

@wcurrangroome I'm convinced there is no purpose for the distinct and that if anything it is removing records we would otherwise want to keep. This drops a very miniscule amount of records which will likely not impact the outcome but still given that I can't find a justification for it I think it's worth removing. My theory actually is that this is filtering out twins or siblings that are very close in age living in the same houshold.

jwalsh28:
> @rpitingolo -- another potential issue: in `0_get_housing_microdata.R`, `distinct()` is applied to microdata. I believe this was because there appeared to be duplicates. However, there were only duplicates (from what I can tell) because identifying attributes were de-selected in a preceding step. So, I think `distinct()` is dropping rows what are in reality unique observations because they happen to share the same subset of attributes that we retain for the analysis. I've commented this out in my re-worked scripts but would appreciate you reviewing and double-checking that this makes sense to you.
> 
> Update: tagging @jwalsh28 here for additional review. JP -- can you take a look at this on the main branch and see if you can deduce why this `distinct()` call was placed there?

@wcurrangroome I'm convinced there is no purpose for the distinct and that if anything it is removing records we would otherwise want to keep. This drops a very miniscule amount of records which will likely not impact the outcome but still given that I can't find a justification for it I think it's worth removing. My theory actually is that this is filtering out twins or siblings that are very close in age living in the same houshold.

cdsolari (approved):
I reviewed the final data output, which matches to where we landed after Erika Tyagi's full review. This is set to what we need, and the final evaluation tests all check out. Erika's branch was also merged in, so all of her data checks are there too. This is approved!

PR #428, opened 2024-11-05
Delete old files and remove medians for affordable and available hous…

awunderground:
…ing for the 30th percent AMI and 50th percent AMI


cdsolari:
FYI, we originally had concerns that our population-weighted medians for all 3 of the AMI levels for affordable & available housing might not have been evaluated for our standards, which is that we needed to be sure that the share of those population-weighted values that were data quality=3 OR NA/missing is <15%. If they were 15%+, then we should hide them from the national medians file. However, after looking, it appears as though they made it, despite having a high rate of missing counties (likely they are all low-population counties). So, we decided NOT to go through with removing any of these from the national medians file.

PR #431, opened 2024-11-25
Update guidance on CT counties and planning regions

malcalakovalski:
Instead of dropping the old CT counties, we decided to keep them and switch to planning regions for data after 2022. This PR updates the documentation accordingly.


malcalakovalski:
@jwalsh28 I replaced the population file with `git checkout version2025 -- geographic-crosswalks/data/county-populations.csv`. Should I do the same for `create-county-populations.qmd`?

PR #433, opened 2024-12-05
Initial political participation metric (voter turnout) updates for 2016 with questions (added as comments in the markdown file)

ridhi96:
In the 2020 update, NJ and KY were flagged as having data quality issues. I am unsure how to interpret the VEST data and figure out which states for 2016 don't have reliable data. I have more concerns around the total votes from the presidential election calculated with the VEST data vs the total votes reported on Wikipedia. 
#408


05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
```diff
@@ -0,0 +1,1020 @@
...
+```
+
+### Quality Checks
+
+#### Compare with MIT
+
+Compare the VEST precinct-level election returns used in our interpolation to the MIT Election Lab precinct-level returns
+
+```{r}
+# QUESTION (Ridhi): Should I remove this section?
```
jwalsh28, line 577: Yes I think this is fine to delete

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
```diff
@@ -0,0 +1,1020 @@
...
+
+if (!dir.exists(mit_directory)) {
+  
+  dir.create(mit_directory)
+  
+}
+
+if (!file.exists(mit_data)) {
+  
+  download.file(
```
jwalsh28, line 618: I'm having trouble with the way R is downloading this file. The CSV that results from this download leads to a fatal error tha makes R abort when I use `read_csv`. Not sure if this is just an issue with my version of R

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
```diff
@@ -0,0 +1,1020 @@
...
+      cv >= 0.05 ~ 2,
+      TRUE ~ 1
+    )
+  )
+
+```
+
+## 2. The process used to create VEST data
+
+QUESTION (Ridhi): There were a lot of precincts/counties where votes were apportioned using weights, boundary conflicts etc. I am not sure how to take this into account for quality here. Below text is for the 2020 metric:
```
jwalsh28, line 911: I looked around at the information VEST provides online and found nothing about significant interpolation happening in 2016 the way they did for the 2020 data due to COVID. Maybe double-check VEST resources to see if you can find anything but I think allocation_quality in 2016 data would all be 1.

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
```diff
@@ -0,0 +1,1020 @@
...
+```{r download-precinct-data, message = FALSE, results = "hide"}
+read_precinct <- function(file) {
+  
+  tempfile <- tempfile()
+  unzip(file, exdir = tempfile)
+  
+  st_read(tempfile) %>%
+    dplyr::mutate(total_votes = across(starts_with("G16PRE")) %>% rowSums) %>%
+    # updated code: added "NAME", "COUNTYFP", "STATEFP" to reflect columns in 2016 data
+    # QUESTION (Ridhi): DO I NEED AN IDENTIFYING COLUMN FOR EACH STATE FILE?
```
jwalsh28, line 182: I don't think you do, it looks like the identifying columns are appended later in the code using spaital joins.

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
```diff
@@ -0,0 +1,1020 @@
...
+# QUESTION (Ridhi): I am not sure what to do about the comments below
+# # Calculate percentage difference between county-level MIT votes and VEST votes aggregated from precincts
+# # XX I'm not sure how useful this is because the results were not identical at the precinct level to begin with
+test_result <- test_result %>%
+   mutate(pct_diff = (abs(total_votes - mit_votes) / ((total_votes + mit_votes)/2)) * 100)
+# 
+# # Kalawao County, HI is missing
+# quantile(test_result$pct_diff, na.rm = TRUE)
+# 
+# # XX Maybe identify counties in the top quartile of percentage difference and flag each place in those counties for quality? Will be annoying because places cross county lines
```
jwalsh28, line 698: This would be one major area for improvement if you have time. I believe the idea here is to compare MIT count-level votes with VEST aggregates and take the top quartile (though play around with what makes sense) of difference in terms of counts. Then identifying which places are in those counties and give them a 3 for quality. This would be an addition to our quality method. Again, I don't think this is crucial but if you have the time  it could be a nice change.

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
```diff
@@ -0,0 +1,1020 @@
...
+# Calculate the areas of the shapefiles for all precincts
+precinct_sf_area <- precincts %>%
+  mutate(precinct_original_area = st_area(.)) %>%
+  select(GEOID16, precinct_original_area) %>%
+  st_drop_geometry()
+
+# Total precinct area is 2,924,296,049,648 m^2
+total_precinct_area <- precinct_sf_area %>%
+  summarize(total_precinct = sum(precinct_original_area)) %>%
+  # QUESTION (Ridhi): What is the purpose of `total_precinct_meters` when area is already in meters squared unit?
```
jwalsh28, line 487: It looks like this step converts the area calculation into square feet (international foot definition (i.e., 1 foot = 0.3048 meter exactly)). I'm not sure what the purpose is but maybe update that variable name to reflect feet instead of meters.

jwalsh28 (approved):
@ridhi96 Overall I think this update is mostly there. A few major items:

1. I put a recommendation for the quality issue (allocation). For 2016 I don't believe there were any major allocation issues like there were in 2020 due to COVID. Double-check this. My suggestion is to set all allocation quality to 1.
2. MIT vs VEST comparison - there is a comment from the 2020 code that suggests flagging counties in the top quartile of difference from the MIT counts. If time permits I think this would be a good addition to our quality method but you would have to make the update to both years.
3. Make sure when you read this file out you update the years to 2016 and please run the evalution function.

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
```diff
@@ -0,0 +1,1020 @@
...
+
+rm(urls, vest_directory, vest_data)
+
+```
+
+Read in the precinct shapefiles and voting results data for each state. Each data set has several variables that begin with `G16PRE` that indicate the number of votes each presidential candidate received in that precinct. We sum these variables to create a new variable called `total_votes`, and only keep this variable and the geometry. We append all states together and create a `GEOID` variable based on row number which we will need for interpolation later.
+
+```{r}
+
+### TO DELETE
```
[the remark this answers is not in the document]
ridhi96: Done

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
```diff
@@ -0,0 +1,1020 @@
...
+# create unique GEOID for interpolation
+precincts <- precincts %>%
+  dplyr::mutate(GEOID16 = row_number())
+
+```
+
+The total number of votes reported in this data is 136,519,876. This is 149,361 votes less than [the national vote total in 2016 of 136,669,237](https://en.wikipedia.org/wiki/2016_United_States_presidential_election#Results_by_state). 
+```{r compare-votes, message = FALSE}
+# total number of votes is 136,519,876  
+# QUESTION (Ridhi): The difference in votes is concerning
```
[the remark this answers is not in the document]
ridhi96: Added this info.

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
ridhi96: I have removed the entire section testing for MIT.

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
ridhi96: Please recheck this! I made some update to this section and the download works for me.

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
ridhi96: I didn't find anything like the interpolation in 2020 although some states like Texas had a lot interpolation based on other factors like change in precinct or other boundaries, reporting of votes for combined geographies etc. I have assigned a quality of 1 to all states for now. More information on data collection is [here](https://dataverse.harvard.edu/file.xhtml?fileId=10596694&version=94.0).

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
ridhi96: Yeah I double-checked that according to the calculation it is converting meters squared into feet squared. I am a little confused as to why this was done because there is a calculation for the area of the US that precincts cover. In the 2020 code, the feet squared number is taken into account for this calculation. When I do a similar calculation I get 2924296049648 / 9328197826286 = 0.313 which is 31.3% and is close to the estimate of 31.46% from the 2020 file. But I think this is wrong as we are dividing what appears to be two different units i.e. feet and meters. But if I take the original area in meters from my data, 9594127956221 / 9328197826286 = 1.02 which is 102%. This also doesn't make sense. I am not sure what to do with this.

ridhi96:
Hi @jwalsh28, I updated the metric based on your review comments. Please take a look at my responses to your comments when you review again. Thank you!

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
jwalsh28: Still getting a fatal error from this that shuts down my R. If it works for you though I think this is probably fine and just a wierd issue with my R version.

jwalsh28 (approved):
Changes look good!

05_local-governance/voter-turnout/voter-turnout-city-2016.Rmd
jwalsh28: Given this is just a test I am not too worried about it. As a note for future reviewers, it is unclear what this test is showing.

jwalsh28 (changes requested):
@ridhi96 Sorry I approved prematurely. To finalize this update can you render `voter-turnout-city-2016.Rmd` so that we have an HTML version and make sure to read out the updated final data.

jwalsh28:
We confirmed with another reviewer that the issue I was having with my R session aborting is likely specific to my software - this is good to go!

jwalsh28 (approved):
Confirmed with another reviewer that the program runs without causing R to abort - good to go!

PR #436, opened 2024-12-13
Iss412

ekgutierrez1:
Backfill SY 2014-2015 through 2018-2019 for cities and counties (total homeless, not subgroup, which starts in 2019-2020)


ekgutierrez1:
One big flag: CT counties are not in crosswalk and therefore not in this data. This will need to be rerun when the crosswalk is fixed.

rpitingolo:
@ekgutierrez1 thanks for the flag. Is this the same crosswalk issue that @wcurrangroome has already figured out? Or is this something else?

ekgutierrez1:
@rpitingolo i use county-populations - so I don't think so!

rpitingolo:
@ekgutierrez1 I finished reviewing and will add line-by-line comments soon. Overall clean intuitive code that runs smoothly. I feel like the comments could use some beefing up. As Stata is not my go-to programming language it took me a while to figure out what was going on in certain places, especially where there are loops operating over multiple variables. I also feel like the quality checks should have more rigid pass/fail criteria. It's difficult for me as someone without topical knowledge to really know whether distributions looks "right" or not. 

More soon. Thanks!

02_housing/Homelessness_metrics_city.do
```diff
@@ -1,15 +1,32 @@
...
+**Housekeeping: install educationdata command **
+cap n ssc install libjson
+net install educationdata, replace from("https://urbaninstitute.github.io/education-data-package-stata/")
+
+*Set up globals and directories
 clear all
 
-global gitfolder "C:\Users\ekgut\OneDrive\Desktop\urban\Github2\mobility-from-poverty"
-global years 2019 2020 2021 // refers to 2019-20 school year through most recent data
+global gitfolder "C:\Users\ekgut\OneDrive\Desktop\urban\Github\mobility-from-poverty"
```
rpitingolo, line 28: Not sure if this is possible in Stata but it would be nice if we didn't have to update the file paths every time someone new runs the code. In R this could be done using .Rproj and relative filepaths for example.

02_housing/Homelessness_metrics_city.do
```diff
@@ -1,15 +1,32 @@
...
+global years 2014 2015 2016 2017 2018
```
rpitingolo, line 29: For someone who isn't familiar with school data it would be helpful to comment or document somewhere whether the year refers to the starting or ending year of the school year (I think it's ending but not 100% certain). I have seen issues with split year data like this before where two programmers weren't on the same page.

02_housing/Homelessness_metrics_city.do
```diff
@@ -80,93 +103,69 @@ foreach year in $years {
...
-According to EdFacts documentation, the variable is used to determine suppression information through 2018-19, 
-and is therefore used in our own suppression determinations for data up through 2018-19. Starting in 2019-20, 
-subgrant_status is no longer used to determine suppression information, and is therefore not necessary for this or future updates.
-*/
-
-*Destring/Create variables needed for data quality check variables 
-foreach var in homeless black hispanic white twomore nh_pi asian amin_an { 
+*create/interpret suppression variables
+	*suppressed observations have between 1 or 2 students, replacing here with 1 so that when aggregated to the city level, we have the best estimate
+foreach var in homeless {
```
rpitingolo, line 168: A suggestion is that for complex loops or code like this it is helpful to have a more detailed comment about what is happening. As I'm less familiar with both Stata and the topical data it took me a while to figure out what is happening here and why.

02_housing/Homelessness_metrics_city.do
```diff
@@ -247,187 +224,119 @@ collapse (sum) *homeless* *black* *hispanic* *other* *white* enrollment , by(yea
...
+		drop _merge
 		
-*check quality
-sum coverage*, d, if homeless_quality==1
-sum coverage*, d, if homeless_quality==2
-sum coverage*, d, if homeless_quality==3
-
-drop enrollment coverage* enroll_* *_districts_suppress min_* count_supp_* *_missing total_race 
+****************
+*Quality Checks
```
rpitingolo, line 272: Is it possible to build quality checks as pass/fail? I'm not sure if this functionality is built in to Stata or if there is a way to create a pass/fail test. I don't see any glaring issues with the check but am also not entirely sure what would be a red flag here.

02_housing/Homelessness_metrics_city.do
```diff
@@ -247,187 +224,119 @@ collapse (sum) *homeless* *black* *hispanic* *other* *white* enrollment , by(yea
...
+	tab check1, m
+	tab check2, m 
+	drop check*
+
+*are all quality flags missing if metric is missing
+tab homeless_quality if homeless_share==.
+tab homeless_quality if homeless_count==.
+
+**************
+*Visual Checks
```
rpitingolo, line 307: Similar comment for visual checks. Useful if you know the topic but otherwise need to be explained what should be checked for in the visual.

rpitingolo (commented):
@ekgutierrez1 left you some comments on the city script. Comments could also apply to the county script. Thanks!

02_housing/Homelessness_metrics_city.do
ekgutierrez1: Apologies, this is the best STATA can do with a global - the nice thing being you only have to change it once at the top.

02_housing/Homelessness_metrics_city.do
ekgutierrez1: Good question - it depends on the source. Data from the education data portal refers to the fall year, while most other sources refer to the spring year. I've added "// these refer to the fall of the school i.e. 2014 = 2014-15" to clarify.

02_housing/Homelessness_metrics_city.do
ekgutierrez1: Apologies, I added some additional descriptions/comments. This loop really is meant for the more recent years of data, where we have data by race/ethnicity in addition to totals.

02_housing/Homelessness_metrics_city.do
ekgutierrez1: Right, its a little subjective depending on the year what might be considered an outlier in the summary statistics. I added a couple "assert" commands where applicable.

02_housing/Homelessness_metrics_city.do
ekgutierrez1: Added (all of these changes apply to both the city and county files)

ekgutierrez1:
The county file for 2014-2018 has been updated with the new crosswalk to include CT for prior years

02_housing/Homelessness_metrics_city.do
rpitingolo: Thanks for adding the clarifying comment.

02_housing/Homelessness_metrics_city.do
rpitingolo: Thanks. It looks like you removed most of the variables from the loop so now it's just looping over a single variable? Is there an intention of adding more later? The code works fine it's just something I notice where there is a loop with only one variable.

02_housing/Homelessness_metrics_city.do
rpitingolo: OK that's helpful. I will have to trust the person(s) looking at the output are knowledgeable enough about the topic to know whether the summary stats and outliers are reasonable or out of bounds. 

Marking this as resolved.

02_housing/Homelessness_metrics_city.do
ekgutierrez1: Right, for these years, race/ethnicity aren't available. The earlier years of the data were somehow since last year, so I took the most recent code (that pulls the race/ethnicity variables) and just removed the subgroup variables so it loops over the 1 variable instead. It felt more efficient at the time than rewriting it all, especially since this specific dofile that pulls older years won't be used again going forward.

rpitingolo: Got it.  Thanks.

rpitingolo (approved):
@ekgutierrez1 thanks!

PR #446, opened 2025-01-13
Fix final evaluation function 

### LINKED ISSUE #445: Fix bug with final data evaluation test function
Function is drawing a bug with multiple years - fix this.

jwalsh28:
Closes #445 

Small tweak to resolve bug in final form evaluation.

### DIFF
--- a/functions/testing/evaluate_final_data.R
+++ b/functions/testing/evaluate_final_data.R
@@ -43,6 +43,9 @@ exp_form_variables <- exp_form %>%
 if(exp_form_variables %>% nrow() > 1) {
   exp_form_variables <- exp_form_variables %>% 
     pivot_wider(names_from = user_input, values_from = c("metric_name_as_written_in_final_data_file", "quality_title", "ci_low_title", "ci_high_title"))
+} else{
+  exp_form_variables <- exp_form_variables %>% 
+    select(-user_input)
 }
 
 
--- a/functions/testing/evaluate_final_data.R
+++ b/functions/testing/evaluate_final_data.R
@@ -61,7 +61,6 @@ expected_subgroups <- exp_form %>%
 expected_years <- exp_form %>% 
   pull(all_years_use_no_space) %>% 
   strsplit(split = ";") %>% 
-  as.numeric() %>% 
   .[[1]]
 
 # check that the file has the necessary columns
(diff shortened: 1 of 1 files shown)


jwalsh28:
@ridhi96 I made the final tweak we discussed on adding in the option to pad the geographies. I believe this does not take away the value of that test as it still checks if there are any cases where the data has a fip with a length that differs from all other geographies in the data, which should never happen.

If you approve after one last look through I will merge with version2025!

ridhi96 (changes requested):
Hi JP,

I am still getting the same error: 

```
Error in `dplyr::mutate()` at functions/testing/evaluate_final_data_4.R:106:3:                                  
ℹ In argument: `place = str_pad(place, width = 5, side = "left", pad = "0")`.
Caused by error:
! object 'place' not found
Run `rlang::last_trace()` to see where the error occurred.
> rlang::last_trace()
<error/dplyr:::mutate_error>
Error in `dplyr::mutate()` at functions/testing/evaluate_final_data_4.R:106:3:
ℹ In argument: `place = str_pad(place, width = 5, side = "left", pad = "0")`.
Caused by error:
! object 'place' not found
---
Backtrace:
     ▆
  1. ├─global evaluate_final_data(...)
  2. │ ├─dplyr::mutate(...) at functions/testing/evaluate_final_data_4.R:106:3
  3. │ └─dplyr:::mutate.data.frame(...)
  4. │   └─dplyr:::mutate_cols(.data, dplyr_quosures(...), by)
  5. │     ├─base::withCallingHandlers(...)
  6. │     └─dplyr:::mutate_col(dots[[i]], data, mask, new_columns)
  7. │       └─mask$eval_all_mutate(quo)
  8. │         └─dplyr (local) eval()
  9. └─stringr::str_pad(place, width = 5, side = "left", pad = "0")
 10.   └─vctrs::vec_size_common(string = string, width = width, pad = pad)
Run rlang::last_trace(drop = FALSE) to see 3 hidden frames.
```

Maybe the error is because of my files. I will investigate this.

Another thing I want to highlight is that the update you made to pad the strings after reading might not catch the case where, for example, a county value incorrectly has length 2 or 1 (chances of this happening are highly unlikely but it could still happen).

jwalsh28:
I see your point on the geography length but I think those cases are a different kind of error. For example, a state that is entered with 1 digit but should have 2 (with the second being a non-zero) would to me represent that the metric someone is producing erroneous FIPs entirely, where length 3 + would say they are including a more extensive FIPs code then required. Harder to correct for the former.

ridhi96 (approved):

PR #448, opened 2025-01-21
First push for code review. Need to fix data quality and CIs

vpancini:

1) This PR relates to #iss422

2) A description of the content in this pull request: 
- **What was changed?** The previous code for this metric was written in Stata. I rewrote it in R and tried to simplify it. Additionally, the previous code was written for a single year at a time, but the rewritten code applies to all years of data (2016-2023) for this metric.
- **What should the reviewer be focusing on?** The reviewer should focus on the construction of the data quality values. When just looking at the 2022 values and comparing them to last year, I couldn't get the values to match despite trying to apply the criteria in several different ways.
- **Is there a logical order to review the files in?** There is only one file (`neonatal-health-all.qmd`)

I can't get the 2022 quality values to match those from last year. If the metric reviewer can see why that might be the case, that would be really helpful. There should be four quality values: 1, 2, 3, or NA. The code from last year didn't apply the quality criteria in a way that made sense to me. E.g., if criteria 1 then quality value A, if criteria 2 then quality value B, if criteria 3 then quality value A again. I tried to take stock of all the criteria for each value and apply the criteria at once, but the quality values ended up not matching. I then thought it was a problem with the hierarchy of the criteria and applied the hierarchy directly from last year instead, but still couldn't get them to match even though the birth rate values did match.


kmartinchek:
1.	Really impressive and efficient code to read in the files from different years. Lines 221-241.
2.	Is there a reason you’re reading in all the years then filtering to 2022? Line 253. Name this in the file if so.
3.	Would also specify why filtering out missing counties and –why—those units are missing. Line 252.
4.	Some code on read births and clean births is somewhat repetitive and could be an opportunity for a function. But it is efficient enough and subgroup formatted varies enough I see why you didn’t.
5.	 Recommend removing files all at the end. Line 310 move to end of file. Simply because they if people want to do back, they have the ability to without re-running code blocks.
6.	What do you mean by line 317? Is that not yet complete?
7.	The results of the full join on line 388 yields a dataset with 3,201 rows. Can you mention if this is expected? This doesn’t match the number of target counties (b/c over time, there are some counties that change codes and are created/removed). It would be good to add validation code to check what you expect vs. what happens in reality.
8.	There are rows with missing “subgroup” fields in county_pop_births_long  what are these fields meant to be?
9.	Can you explain more what you are doing in the block of data following line 405? I am confused what you are aiming to do and it makes it hard to assess the code (which seems to be filling in missing data from unidentified counties with state values – but I am not sure of the intuition as to why you’re doing this fill-in method). Wouldn’t you want to proportionally distribute based on county population or some other fill-in strategy to not double-count births?
10.	Should line 441 be an & or an | in the if_else statement? 
11.	Should the suppressed and unidentified county flags be generated separately for overall births and low birth weight? They seem to vary (where low birth weight is more likely to be suppressed, for example).
12.	Code starting on line 458 gives a warning that NAs are introduced by coercion – is that what you want?
13.	I see you suppress data for those where the numerator is less than 30. (On line 472, you state “denominator”, I don’t think that is right terminology). However, I think it would be good to note how you treat 0 low birth weight observations. They are listed as 0 percent – but I think should be missing based on statement in line 472.
14.	Also, in this section following line 472, you suppress when there are fewer than 30 low birth weight observations, but I see some observations with 0 low birth weight observations and fewer than 30 overall births that are not NA but 0s. I would reassess this. I think that if the code in this section isn’t what you anticipate, it is hard for me to assess the application of the quality flags. Especially the ratio-based adjustments.
15.	Line 494 complicates what I am looking at—is it the numerator or denominator that should be more than 30? Overall, can you give a comprehensive description of how you are determining quality? There are some contradictions here that is making it hard for me to assess what we want quality to be and what we get.
16.	 Shouldn’t line 515 be births_all < 100 & births_all >= 30 ~ 3? Based on description in line 494.
17.	I’m not sure the order of these complex conditions is right. I would define quality 1 last. Also, you overwrite the dataset and variables when defining quality. I would recommend when debugging splitting this out and then removing intermediate variables so you can better see where the error is emerging. Similarly, one strategy would be initialize the values to be something like 9 so you can see when quality conditions are not applied to observations (line 507).
18.	What is the intuition of line 517 and 519?
19.	When you check quality distribution on line 539, please make it clear what we expect so we can examine accordingly.
20.	Once I have more clarity on quality flags above, will take another look at debugging.
21.	I see the CIs are in-progress and did not do a detailed review. But line 656 has an error (because the test data is not being piped into the select() and rest of pipeline. Anticipate this is an expected error.

cdsolari:
@vpancini In just looking at the output files, it seems like the estimates were rounded a bit early. We should not round, as it grants us flexibility in what we show. It also strikes me that we have similar-looking estimates within states, more than I'd expect. I tested for 2022, and it looks like it just appears that way because you round to the hundredths place in the all county file, but remember that we display a value of 0.111 as 11.1%, and the way you have it would lose the needed decimal place. Regardless, please do not round. You calculated years that were already calculated, which is fine, except that if you do that, you should then compare your values for re-calculated years with past rounds of data calculated for the same year. If you had done that, none would match because of the rounding.  

In the subgroup file, I also noticed that you rounded to the thousandths place which is better, but still should not be done. Also, for the subgroup-type "all", the corresponding subgroup should take on the value "All", but instead it has "NA".

kmartinchek:
I ran the code without errors and looked at code logic, including on the new CIs and quality variables. Really great job addressing comments and I learned a few little tricks in wrangling in R from it. Based on this review, this looks good. I just do see that there is a lot of missingness and low quality in the final extract, but I think this was also true for past updates and aligns with expectations.

I have not checked this output against the old updates to make sure it matches, but would recommend doing so as an extra check because this code was re-written in this update. I think also because this code is in R, it needs the verification function? At least having a check that the data has the expected number of observations for each subgroup would be helpful at the end.

vpancini:
Responding to Kassie's comments:
1. Thank you :)
2. Was just trying to recreate the values from 2022 (last year) first since I re-wrote the code in R. Removed this filter.
3. Done
4. Leaving the repetitive code this round
5. Fair enough, removed
6. Line 317 was "todo(): The 2022 counties for Connecticut don't match between the CDC WONDER data and Urban's county population file." - we know they don't match and aren't going to try and crosswalk them this year
7. This is mostly due to the unidentified counties in the CDC WONDER data. I added code to show the level of missingness we expect from each file at the join
8. Those are the "all" or total values (non-subgroup). I updated to assign a value of "all" to these observations
9. This is a good point. I was just doing what I inherited from previous code, and there was nothing in the Git issue about updating this strategy, but I agree that there is probably a better method. I included it in the "Considerations for future updates" section of my code and will flag it in my Git issue as well.
10. It doesn't matter. The "Unidentified Counties" label is based on total county population, so it will apply to all births and lbw births. I just included both out of preference. It could have been one or the other, or both. The result should be the same.
11. This was unexpected, but I fixed it earlier in the code. It was from counties in CDC labelled "Missing County" that, after 2013, got incorporated into the "Unidentified Counties" group in their states. Not sure why CDC still includes these observations anyway, but I dropped them earlier in the code
12. I think the rest of the comments were mostly related to my own confusion around the data quality values, which have been updated. Please reach out to me on Slack if you have any questions as you re-review or if anything is still unclear!

kmartinchek:
> Responding to Kassie's comments:
> 
> 1. Thank you :)
> 2. Was just trying to recreate the values from 2022 (last year) first since I re-wrote the code in R. Removed this filter.
> 3. Done
> 4. Leaving the repetitive code this round
> 5. Fair enough, removed
> 6. Line 317 was "todo(): The 2022 counties for Connecticut don't match between the CDC WONDER data and Urban's county population file." - we know they don't match and aren't going to try and crosswalk them this year
> 7. This is mostly due to the unidentified counties in the CDC WONDER data. I added code to show the level of missingness we expect from each file at the join
> 8. Those are the "all" or total values (non-subgroup). I updated to assign a value of "all" to these observations
> 9. This is a good point. I was just doing what I inherited from previous code, and there was nothing in the Git issue about updating this strategy, but I agree that there is probably a better method. I included it in the "Considerations for future updates" section of my code and will flag it in my Git issue as well.
> 10. It doesn't matter. The "Unidentified Counties" label is based on total county population, so it will apply to all births and lbw births. I just included both out of preference. It could have been one or the other, or both. The result should be the same.
> 11. This was unexpected, but I fixed it earlier in the code. It was from counties in CDC labelled "Missing County" that, after 2013, got incorporated into the "Unidentified Counties" group in their states. Not sure why CDC still includes these observations anyway, but I dropped them earlier in the code
> 12. I think the rest of the comments were mostly related to my own confusion around the data quality values, which have been updated. Please reach out to me on Slack if you have any questions as you re-review or if anything is still unclear!

I just finished the re-review and think things look good from a code logic POV-- I really appreciated the extra comments and they were perfect for understanding what you were doing, why, and expected result so I could check my understanding too. I added a comment above about the re-review.

cdsolari:
@kmartinchek and @vpancini I already checked the missingness compared to past rounds as well. It is as we would expect. Thank you!

cdsolari (approved):
The final evaluation form was added and all tests passed. The final files look good too. This has been through several reviews, and this is now clean and final.

PR #450, opened 2025-01-28
Update college metric. 

jwalsh28:
Back-fill yeras 2014 and 2016. Apply new AWS d…ata storing approach. Bring in final evaluation tests.

1) A link to the issue that this PR relates to. You can bring up a list of suggested issues and pull requests within the repository by typing.

#403 

2) A description of the content in this pull request.  
- What was changed?
I back-updated for 2014 and 2016. The format of the program has changed to make it easier to update individual years and run validation and tests collectively at the end. This program also now utlizes AWS for intermediate file storage.
- What should the reviewer be focusing on?
The AWS step is new and the biggest conscern is that this file storage would introduce some threat to the data content.
- Is there a logical order to review the files in?
Look at the API extract functions first:
1) extract_ipums_aws.R
2)extract_repwt_college_aws.R
Then I would start with the place file college_metrics_place.qmd

You will need to have access to the AWS keys for this review to function best. Please slack or email me and I will provide those.


08_education/college_metrics_county.qmd
```diff
@@ -947,8 +1033,72 @@ metrics_education_sub <- metrics_education_sub %>%
...
+```
+
+If updating a single year the program ends here. To see validation results re-run with selection that all years of intermediate data are available.
+
+```{r}
+if(single_year_update == "Yes"){
+  knitr::knit_exit()
+}
+```
+
```
malcalakovalski, line 1065: This is such a good idea, I didn't know about this function!

08_education/run_college.qmd
```diff
@@ -0,0 +1,88 @@
...
+
+reports <-
+  tibble(
+    input = here::here("08_education", "college_metrics_place.qmd"),
+    output_file = glue::glue("calculate-college-county-{years}.html"),
+    execute_params = map(years, ~ list(year = .))
+  )
+
+pwalk(reports, quarto_render_move)
+```
```
malcalakovalski, line 88: This code is awesome and makes the code so reproducible and extendible!

functions/API/extract_ipums_aws.R
```diff
@@ -0,0 +1,132 @@
...
+      read_ipums_micro(
+        ddi,
+        data_file = here(folder_path, extract_gz_filename)
+      )
+    
+    #DDI is a codebook that is used by IPUMSR to format the micro data downloaded
+    #Lower variable names and get rid of unnecessary variables
+    acs_imported <- micro_data %>%
+      rename_with(tolower) %>% 
+      select(-serial, -raced, -strata, - cluster, -hispand, -empstatd)
```
malcalakovalski, line 100:
```suggestion
      select(-serial, -raced, -strata, -cluster, -hispand, -empstatd)
```

functions/API/extract_ipums_aws.R
```diff
@@ -0,0 +1,132 @@
...
+    
+    rm(micro_data)
+    
+    #Zap labels and reformat State and PUMA variable
+    acs_imported <- acs_imported %>%
+      mutate(  
+        across(c(sample, gq, race, hispan), ~as_factor(.x)),
+        across(c(sample, gq, race, hispan, sex, diffcare, diffsens, diffmob, diffphys, diffrem), ~as_factor(.x)),
```
malcalakovalski, line 108: Why are `sample`, `gq`, `race`, and `hispan` converted to factor twice? Should only line 108 be kept?

malcalakovalski:
- Overarching thoughts
  - Excellent documentation
    - If possible, I would add a bit more on the why for beginners if that feels relevant
    - For example, why are replicate weights necessary? I think linking to somewhere where a collaborator could find more info would be great
  - Code is modular and efficient
    -  `extract_ipums_aws.R` and `extract_repwt_college_aws.R` are really really excellent. I think this modularity makes the `.qmd` files so much more readable. Also, saving these ipums extracts on S3 is a great call to avoid repeat downloads and faster code review!
    - One room for improvement is that a lot of the plots are similar and could be converted to a function (e.g., the disability variable subgroup plots)
    - Similarily, when final metrics were created for subgroups, the code was a bit copy paste and the logic could be extracted into a function. I think it wouldn't sacrifice readability either. However, this is more of a "nice to have" suggestion which could be left for a future update. 
    - I'd suggest breaking it into three functions:
      - `calculate_effective_sample`
      - `summarise_effective_sample`
      - `add_suppressed_rows`
  - I'd consider using tidylog to track observations dropped and joins
  - The summary tables at the end confirming the number of rows matches to the expected counties/places but also showing the counts and shares of suppressed rows are extremely valuable
  - I don't see any issues at all with the AWS S3 usage. Was the concern that the encoding of the data would be different or there might be differences between locally saved files vs the cloud? I think the code working and the files matching expectation are good indicators that's not the case . As a sanity check, there could be value in saving locally as well and then testing that the files are equal.

malcalakovalski (approved):
- Overarching thoughts
  - Excellent documentation
    - If possible, I would add a bit more on the why for beginners if that feels relevant
    - For example, why are replicate weights necessary? I think linking to somewhere where a collaborator could find more info would be great
  - Code is modular and efficient
  - `extract_ipums_aws.R` and `extract_repwt_college_aws.R` are really really excellent. I think this modularity makes the qmd files so much more readable. Also, saving these ipums extracts on S3 is a great call to avoid repeat downloads and faster code review!
    - One room for improvement is that a lot of the plots are similar and could be converted to a function (e.g., the disability variable subgroup plots)
    - Similarily, when final metrics were created for subgroups, the code was a bit copy paste and the logic could be extracted into a function. I think it wouldn't sacrifice readability either. However, this is more of a "nice to have" suggestion which could be left for a future update. 
    - I'd suggest breaking it into three functions:
      - `calculate_effective_sample`
      - `summarise_effective_sample`
      - `add_suppressed_rows`
  - I'd consider using `tidylog` to track observations dropped and joins
  - The summary tables at the end confirming the number of rows matches to the expected counties/places but also showing the counts and shares of suppressed rows are extremely valuable
  - I don't see any issues at all with the AWS S3 usage. Was the concern that the encoding of the data would be different or there might be differences between locally saved files vs the cloud? I think the code working and the files matching expectation are good indicators that's not the case . As a sanity check, there could be value in saving locally as well and then testing that the files are equal.

jwalsh28:
Closes #403

PR #451, opened 2025-01-30
Update to employment metric.

jwalsh28:
**Mobility metric pull request template**

Please include the following points in your PR:

1) A link to the issue that this PR relates to. You can bring up a list of suggested issues and pull requests within the repository by typing #.

Closes #419

2) A description of the content in this pull request.  
- What was changed?
- What should the reviewer be focusing on?
- Is there a logical order to review the files in?

3) Detail on any issues or flags that the metric reviewer/data-team should be aware of.

PR #456, opened 2025-02-04
Update to income metric.

jwalsh28:


malcalakovalski:
I have similar overall comments to the PR's for issues #450 and #451. However, one thing I noticed is missing here is running the the files through the final evaluation function. The subgroup evaluation form csvs are also missing from `10a_final-evaluation`

01_financial-well-being/R/run_income_years.qmd
```diff
@@ -0,0 +1,99 @@
+---
+title: "Run employment"
```
malcalakovalski, line 2:
```suggestion
title: "Run income"
```

malcalakovalski (changes requested):

jwalsh28:
@malcalakovalski Apologize this intitial PR was a bit half-baked and did not include subgroups - they have been added. As well as the final evaluations. You should be good to proceed with your review.

malcalakovalski (approved):
Everything looks great except that I would incorporate the logic of `functions/testing/evaluate_final_data_srv_quantile.R` into the `evaluate_final_data` function used across the project. The new function is the same except that it allows for NA values for lb and ub in confidence intervals. However, this could be handled by adding an argument for whether the data is survey data or not in the function and do `if else` logic based on that. That's okay for this round, but I'm flagging this for the future!

PR #460, opened 2025-02-05
Add stata code for final eval function

jwalsh28:
This code is the STATA version of the final evaluation function that was originally written in R.


cdsolari (approved):
This is Manu's code based on his GitHub CoPilot of the R-script to test the metric lead's output against the final evaluation forms.

PR #461, opened 2025-02-05
Update tract to place crosswalk file to use 2014 places from geocorr

jwalsh28:
Replaces the geocorr2018_tract_to_place file with an updated CSV that includes 2014 places instead of 2010 places.

The table below shows the match rate with our target geography place list. You can see that the old crosswalk has three missing places while the new crosswalk only has one. This is expected because of South fulton county city.

cw_old cw_new     n
  <chr>  <chr>  <int>
1 old    new      483
2 NA     new        2
3 NA     NA         1


awunderground (changes requested):
I'm concerned about changing the name of the current crosswalk. This will break the air quality metric. 

I recommend adding the new file with a new name like geocorr2018_2010tract_to_2014place.csv

jwalsh28:
@awunderground Point well taken, I returned the name of the historical file and went with your suggestion for the updated crosswalk

awunderground (approved):

PR #484, opened 2025-03-28
Iss483 medians

awunderground:
This PR updates the medians code to work with the new mobility metrics. It also adds 2014, which wasn't previously included. Finally, it deletes a few files that are no longer needed.


jwalsh28 (approved):
The visuals in this comparing quality are so insightful, this is a very valuable document to have. Looks great!

PR #488, opened 2025-04-25
Suppress select values

### LINKED ISSUE #487: Data suppression for weak or impluasible values

jwalsh28:
Closes #487 

Adds suppression of select variables to the combine data files for place overall and race/ethnicity.


cdsolari (commented):
I think we need to extend the suppressions around crime rates to the juvenile arrest case as well. And, I think we have the most confidence in the 2023 data, so the suppression can be for just pre-2023.

cdsolari (approved):
Looks like the county and city changes we expected. Looks good!

jwalsh28:
Closes #488

PR #490, opened 2025-04-30
Change county metric to statistics calculated in the subgroup document.

awunderground:
- Deletes the county transportation .qmd
- Uses the county subgroup .qmd to create the county-level transportation metrics


jwalsh28 (approved):
This fix looks good. The new `transportation_all_county.csv` now uses the "All" values calculated from the subgroup file (up from the tract-level). Besides that the file format was kept the same which is what we want.

PR #521, opened 2025-06-20
Evaluation function updates

ridhi96:
Updated the evaluation function to include multiple checks on the metrics using the Technical Specification with error handling.

- The workflow assumes that a final data frame will have columns and data aligned with the expectations in the Wiki (for both overall and subgroups csv files). 
- When the `evaluate_final_data()` function is run on final data files generated during version 2025 update, not all files will pass the checks because earlier the Metric Leads had more freedom to structure the data files. It would be helpful to evaluate the checks that fail while referencing the data file's content.
- This upgrade aims to move away from final evaluation forms for each metric to consolidate data specification.
- Based on the review and nuances of certain metrics, the code might need further updates.
- Right now the code expects the final data file to have all metric years present. This is going to be re-evaluated and updated after initial review.


.github/ISSUE_TEMPLATE/mobility-metric-update-issue-template.md
```diff
@@ -1,56 +1,78 @@
 ---
-name: Moibility Metric Update Issue Template
+name: Mobility Metric Update Issue Template
```
jwalsh28, line 2: Good thing to have accurate spelling in the QC code ...

data/technical_specification/metrics_technical_spec_2026.csv
```diff
@@ -0,0 +1,90 @@
+﻿METRIC_ID,METRIC,GEOGRAPHY,YEARS,SUBGROUP_TYPE,SUBGROUP,SUBGROUP_YEARS,CI_LOWER,CI_UPPER,HAS_CI,QUALITY,RANGE,OVERALL_DIFFERENT,FINAL_DATA_PATH
```
jwalsh28, line 1: @ridhi96 This file is great. Did you make it manually or programmatically? If there is code behind it can we also uplaod that to the repo somehwere.

data/technical_specification/metrics_technical_spec_2026.csv
```diff
@@ -0,0 +1,90 @@
...
+1,share_employed,county,2023; 2022; 2021; 2019; 2018; 2017; 2016; 2015; 2014,all,All,2023; 2021; 2018; 2016; 2014,share_employed_lb,share_employed_ub,Yes,share_employed_quality,0-1,Yes,
```
jwalsh28, line 2: Do we have an equivalent file for place level data?

functions/testing/evaluate_final_data_checks.R
```diff
@@ -0,0 +1,531 @@
...
+#' @param confidence_intervals  (logical): A TRUE or FALSE value indicating if the final data has confidence intervals.
+#' 
+#' @return: A series of test outcomes with a summary of errors encountered.
+#' 
+#' @example evaluate_final_data(
+#'                              metric_var = "share_digital_access", 
+#'                              data = df_final_data, 
+#'                              geo = "county", 
+#'                              subgroups = TRUE, 
+#'                              subgrp_type = c("all", "race-ethnicity", "income"),
```
jwalsh28, line 38: With the way this sets up the subgroup_type argument am I right to believe that a metric lead could still choose to read out and evalutae individaul subgroup files for the metric separately? Or do they have to test a combined file?

jwalsh28: In addition, will they always have to include "all" to have it test the all rows of a subgroup file? Should we consider making this the default?

functions/testing/evaluate_final_data_checks.R
```diff
@@ -0,0 +1,531 @@
...
+  # Check if metric (and subgroup types) exists in technical specification
+  check_metric_subgroup_exists(metric_var, tech_spec_all, geo, subgrp_type)
+  
+  # Tech spec for relevant metric
+  tech_spec <- get_subset_tech_spec(tech_spec_all, metric_var, geo, subgroups)
+  
+  if ( (isTRUE(subgroups)) & (!is.null(subgrp_type)) ) {
+    
+    # Check user provided subgroup types against subgroup types in data
+    check_user_data_subgroup_types(data, subgrp_type)
```
jwalsh28, line 64: This will fail if the user doe not provide "all" as a subgroup type - which is fine, I think we just need to make it clear to be included or make "all" a default.

functions/testing/evaluate_final_data_checks.R
```diff
@@ -0,0 +1,531 @@
...
+  # Pull year list for overall or each subgroup type from technical specification
+  expected_years <- get_expected_years(tech_spec, subgrp_type)
+  
+
+  # Pull expected variable names
+  expected_variables <- get_expected_metric_vars(tech_spec)
+  
+  ### CHECKS
+  
+  checks <- list(
```
jwalsh28, line 80: This is an awesome way to run all of the checks - big programmer brain!

functions/testing/evaluate_final_data_checks.R
```diff
@@ -0,0 +1,531 @@
...
+#' Checks if required years are present 
+#' 
+#' @param data The data that is staged to be read out as the final file.
+#' @param exp_years A data frame of subgroup types and their years, or a unique vector 
+#' of years if no subgroup type is provided.
+#' @param subgroups A TRUE or FALSE value indicating if the final data has subgroups.
+#' 
+#' @returns NULL (throws error if incorrect years are present for metric and/or subgroups)
+#'  
+check_expected_years <- function(data, exp_years, subgroups) {
```
jwalsh28, line 528: This will be contigent on the tech file getting updated appropriately as the year variable is the only thing likely to change consistently moving forward. Might be worth adding to the failure note something like "Has the tech file been updated to reflect the latest data?"

jwalsh28 (changes requested):
This is an excellent improvement on the original process.

There is good flexibility built into this, it gives metric leads the ability to run the test on individual subgroup output. The only lock is around expected years, so programs that are setup to read out individual years will have to adjust. We may want to consider providing flexibility here.

My other note is I think the way this considers "all" is a bit confusing right now. Can we consider an option to include "all" as a default without the metric lead having to include it in the subgrp_type argument?

data/technical_specification/metrics_technical_spec_2026.csv
ridhi96: I created the file manually with some help from Claudia. We do need to update the place data in this file.

data/technical_specification/metrics_technical_spec_2026.csv
ridhi96: Not yet.

functions/testing/evaluate_final_data_checks.R
ridhi96: I updated this code so default is to test just for "all" subgroup if user doesn't mention it. Otherwise the user will have to mention something like c("all", "income") if they need to test multiple subgroups with the "all" subgroup type.

functions/testing/evaluate_final_data_checks.R
ridhi96: This now lets the metric lead indicate whether the file contains all expected years or a subset of them. If only a subset are provided, then the information is flagged to the metric lead about which years are missing but the data will still pass all tests. We'd have to rely on the metric lead to provide final data for all relevant years.

ridhi96:
Hi @jwalsh28, thank you for the review! I have updated the code based on your suggestions so that now subgroup type checks defaults to "all" and specific metric years can be tested.

jwalsh28:
The most significant transition that this PR makes is to use a main "technical_spec" file to base the evaluation tests on, as opposed to our current process of indepedent "final evaluation forms".  I think this goal is lofty but a good one. We could include it as a goal in this next tranche3 update - but it currently is note ready to be merged.

jwalsh28 (approved):
Lets merge!

PR #533, opened 2025-10-15
Update descriptive representation

jwalsh28:
This PR brings in the 2024 pep data to our geographic crosswalks


malcalakovalski (approved):
This is reproducible and in great shape! I only have two very minor comments:

The `tryCatch` logic in `read_places` and `read_counties` is helpful but could be further modularized since only the `survey` parameter changes. I would suggest something like this:

```
read_places <- function(update_years) {
  avail <- tryCatch(
    load_variables(update_years, "acs5", cache = TRUE),
    error = function(e) e
  )

  survey <- if (inherits(avail, "error")) "acs1" else "acs5"

  vars <- c(
    total_people       = "B01003_001E",
    total_hispanic     = "B03003_003E",
    total_nonhisp      = "B03002_002E",
    white_nonhispanic  = "B03002_003E",
    black_nonhispanic  = "B03002_004E",
    aian_nh            = "B03002_005E",
    asian_nonhispanic  = "B03002_006E",
    nhpi_nh            = "B03002_007E",
    other_nh           = "B03002_008E",
    two_or_more_nh     = "B03002_009E"
  )

  get_acs(
    geography = "place",
    year = update_years,
    survey = survey,
    variables = vars,
    output = "wide"
  ) %>% 
    dplyr::mutate(year = update_years)
}
```

I'm also wondering why the latest year (2024) wasn't chosen for the distribution of population share visualizations? Was that to ensure the 2023 visualizations from last round remained the same? 

In any case, these are very minor suggestions/questions and I think this PR is good to be merged as is.

awunderground:
Can we migrate this to `22_descriptive-representation/`?

jwalsh28:
> Can we migrate this to `22_descriptive-representation/`?

Done!

PR #538, opened 2025-11-04
Revert changes to desc rep back to 5-year ACS system

jwalsh28:
Remove addition of 1-year acs file


malcalakovalski (approved):
Changes all look good!

PR #562, opened 2026-01-26
Update employment 1-year data for 2024

jwalsh28:
This PR updates the employment metric with 2024 data

Main changes are:

1) Update the dates in the QMD files to read in 2024
2) Fix an issue with the IPUMS pull function because of a changed variable name
3) Create a subfolder to organize single year HTLM output files


malcalakovalski (approved):
This looks good to me! I do think 90% of the code seems similar across `.qmd` files and I wonder if this metric could benefit from a single parametrized quarto doc. Alternatively, the common logic across the docs could be extracted into functions in `09_employment/R/` (data cleaning, validation, etc.). 

Also, out of curiosity, what model did you use for the AI documentation? 

I also think using `urbnthemes` for consistency with the rest of the plots in this project would be good!

PR #565, opened 2026-02-05
Update housing affordability metric with 2024 data    

malcalakovalski:
This PR updates the housing affordability metric with 2024 data                                                       
                                                                                                                        
  Main changes are:                                                                                                     
                                                                                                                        
  - Update the dates in the QMD file to include 2024                                                                    
  - Fix an issue with the FIPS variable rename that only worked for 2023                                                
  - Fix the geography count test for places by adjusting for South Fulton, GA (not in pre-2022 crosswalk)               
  - Remove unnecessary county adjustment for CT county-equivalents                                                      
  - Move files to `11_housing-affordability/` folder   
  - Simplify quality check comparison file to only compare final data between versions (the methodological variant comparisons were a one-time exploration and don't need to be reproduced)


11_housing-affordability/ratio_housing_affordable_available.qmd
```diff
@@ -0,0 +1,1932 @@
...
+
+if (!dir.exists(temporary_data_path)) {
+  dir.create(temporary_data_path, recursive = TRUE)
+}
+```
+
+## Step 1: Download microdata
+
+-   Submit queries to the API, separately, for both occupied and vacant housing unit data
+-   Note that this will take approximately one hour if you have not downloaded these data previously.
```
jwalsh28, line 179: We may want to consider using the AWS storage method to make reveiwing this/updating this less time intensive for internal users. See https://github.com/UI-Research/mobility-from-poverty/blob/main/08_education/college_metrics_county.qmd as an example of the process.

11_housing-affordability/ratio_housing_affordable_available.qmd
```diff
@@ -0,0 +1,1932 @@
...
+      statefip = sprintf("%0.2d", as.numeric(statefip)),
+      puma = sprintf("%0.5d", as.numeric(puma))
+    ) %>%
+    arrange(statefip, puma) %>%
+    # save temp file with API pull
+    write_csv(here(temporary_data_path, paste0("housing_microdata_", year, ".csv")))
+}
+
+# Helper function to check for the presence of already-downloaded microdata
+get_years_to_query <- function(
```
jwalsh28, line 277: I believe that the IPUMSR functionality allows to download one file with multiple 1-year ACS pulls. Flagging as a possible smoother download process instead of individual 1-year files.

11_housing-affordability/ratio_housing_affordable_available.qmd
```diff
@@ -0,0 +1,1932 @@
...
+    distinct_counties = n_distinct(countyfip)
+  ) %>%
+  left_join(
+    county_populations %>%
+      group_by(year) %>%
+      summarize(distinct_counties_benchmark = n_distinct(county_fip))
+  ) %>%
+  mutate(
+    distinct_counties_benchmark = case_when(
+      year < 2022 ~ distinct_counties_benchmark + 8, ## addressing missing CT counties
```
jwalsh28: I think this is obsolete after changes made to the county populations file that include the CT counties, it results in the test failing. Consider removing the test or updating the logic.

11_housing-affordability/ratio_housing_affordable_available.qmd
```diff
@@ -0,0 +1,1932 @@
...
+
+## Some place names are missing (and corresponding place codes are effectively NA)
+## This is fine. These are where counties don't overlap with places. We retain these
+## because there are a few (n = ~10) counties that don't overlap with any places,
+## and if we drop these, we can't benchmark to our total expected county population
+## numbers as easily/accurately
+# county_place_crosswalk_longitudinal %>%
+#   skimr::skim()
+
+write_csv(county_place_crosswalk_longitudinal, here("geographic-crosswalks", "data", "county_place_crosswalk_longitudinal.csv"))
```
jwalsh28: This is the only metric code that I know of that directly creates a geographic crosswalk file, I'm a bit lost at what this file even does. We should revisit the need of this section and either update our crosswalk folder directly or transition to exsiting files in the folder.

11_housing-affordability/ratio_housing_affordable_available.qmd
```diff
@@ -0,0 +1,1932 @@
...
+          .cols = c(RENT, RENTGRS),
+          .fns = ~ if_else(.x == 0, NA, .x)
+        ),
+        year = str_extract(.x, "[0-9]{4}"),
+        VALUEH = if_else(VALUEH == 9999999, NA, VALUEH),
+        crosswalk_period = as.character(crosswalk_period)
+      ) %>%
+      filter(
+        PERNUM == 1, # one observation per household
+        GQ < 3
```
jwalsh28, line 558: We should be able to add this filter into the IPUMSR pull directly (see note above)

11_housing-affordability/ratio_housing_affordable_available.qmd
```diff
@@ -0,0 +1,1932 @@
...
+    samples = c(paste0("us", year, "a")), # use ACS data
+    variables = c(
+      "HHWT", ## household weight -- appropriate for statements about households
+      "PERWT", ## person weight -- appropriate for statements about individuals
+      "REPWT", ## replicate weights
+      "PERNUM", ## unique individual
+      "ADJUST", ## a factor for adjusting within-calendar-year costs
+      "STATEFIP", ## 2-character state FIPS code
+      "PUMA", ## code identifying the public-use microdata area (PUMA)
+      "GQ", ## "general quarters"--identifies the household type
```
jwalsh28: I'm fairly sure this code filters out instituionalized groups, we can do that here to save some space on the data download.

11_housing-affordability/ratio_housing_affordable_available.html
jwalsh28, line 1: Is this outdated? Does not seem to be the ouput of the latest version of the QMD document "11_housing-affordability/ratio_housing_affordable_available.qmd"

11_housing-affordability/ratio_housing_affordable_available.qmd
```diff
@@ -0,0 +1,1932 @@
...
+## specific links used in the code will likely need to be updated year-to-year
+
+if (geography == "place") {
+  metrics_prior1 <- read_csv(
+    ## this corresponds to longitudinal, place-level metrics that are broken down by tenure
+    "https://urban-data-catalog.s3.amazonaws.com/drupal-root-live/2024/09/67_mobility-metrics_place_tenure_longitudinal.csv"
+  )
+} else {
+  metrics_prior1 <- read_csv(
+    ## this corresponds to longitudinal, county-level metrics that are broken down by tenure
```
jwalsh28, line 1675: Should we update this to compare with the 2025 data?

11_housing-affordability/ratio_housing_affordable_available.qmd
```diff
@@ -0,0 +1,1932 @@
...
+  mutate(
+    name = name %>% str_replace_all("_", " ") %>% str_to_sentence() %>% str_remove(" difference flag"),
+    percent = value / count
+  ) %>%
+  ggplot() +
+  geom_col(aes(x = name, y = percent, fill = subgroup), position = "dodge") +
+  coord_flip() +
+  scale_y_continuous(labels = scales::percent_format()) +
+  labs(
+    title = "Share of non-missing geographies with differences in calculated values for 2022 data, 2024 vs. 2023" %>% str_wrap(90),
```
jwalsh28: Again here I think we are comparing 2025 and 2024 - we could also update the 2022 comparsion to 2023

11_housing-affordability/ratio_housing_affordable_available.qmd
```diff
@@ -0,0 +1,1932 @@
...
+        filter(if_any(.cols = matches("difference"), .fns = ~ .x < 0)) %>%
+        nrow() == 0
+    )
+  }
+)
+```
+
+***There Are No Breaking Changes to Prior Years' Estimates***
+
+Unless substantive changes have been made to the process of calculating the metrics, every metric value for every geography calculated during the current cycle should be exactly the same as the value calculated during the preceding year.
```
jwalsh28, line 1660: These results have not changed from the last update but I am trying to understand how to interpret this. Do the large shares of counties with different rent results for 2022 indicate that the methodology changed sigfinicantly in 2024?

11_housing-affordability/ratio_housing_affordable_available.qmd
```diff
@@ -0,0 +1,1932 @@
...
+    facet_wrap(~name) +
+    urbnthemes::theme_urbn_print() +
+    labs(
+      x = "Percent annual change in metric",
+      y = "Count",
+      title = "Metrics should show relatively little year-over-year variation with distributions clustered around 0%",
+      subtitle = "Distributions of year-over-year changes in housing affordability and availability metrics, by AMI level and subgroup"
+    ) +
+    scale_y_continuous(labels = scales::comma) +
+    scale_x_continuous(labels = scales::percent)
```
jwalsh28: I think this graphic would be more informative if we set limits for the X axis so that you can see the YOY change numbers more clearly

jwalsh28 (commented):
I think we need to update a few of the tests to reflect the latest year.

Biggest priority is test starting at line `1605` which looks at changes from prior years' estimates. I believe this is still pulling the 2024 version of the metrics, we now want to be comparing with the 2025 version. My recollection is that the methods changed a lot in 2024 which shows up as signficant county differences for the rent metric. We would anticipate not to see these when comparing with 2025.

11_housing-affordability/ratio_housing_affordable_available.qmd
malcalakovalski: I agree, this is obsolete. I will remove it!

11_housing-affordability/ratio_housing_affordable_available.qmd
malcalakovalski: I think this is redundant so I will refactor to just using the existing crosswalks

11_housing-affordability/ratio_housing_affordable_available.qmd
malcalakovalski: You're totally right, I switched to pulling multiple years and also filtering the data directly like you suggested. This sped up the API call significantly.

11_housing-affordability/ratio_housing_affordable_available.html
malcalakovalski: Re-rendered!

11_housing-affordability/ratio_housing_affordable_available.qmd
malcalakovalski: Yes, I updated this too!

11_housing-affordability/ratio_housing_affordable_available.qmd
malcalakovalski: Yes, your interpretation seems right to me. The large renter differences in the 2024 vs. 2023 comparison would reflect a methodology change made in that update. Since the methodology didn't change this round the current comparison of 2025 vs. 2026 shows zero differences

11_housing-affordability/ratio_housing_affordable_available.qmd
malcalakovalski: Great call, I set limits to zoom in on those numbers!

11_housing-affordability/ratio_housing_affordable_available.qmd
malcalakovalski: Great call! I added AWS caching following your example

jwalsh28 (approved):
These changes look great Manu, I think this is all set!

PR #568, opened 2026-02-09
Convert homeless students metric to R/Quarto and consolidate 2014-2022 data

malcalakovalski:
## Summary

This PR updates the homeless students metric (Housing Stability predictor) by converting the Stata workflow to R/Quarto and consolidating data processing into a single parameterized document.

1. Converted Stata to R/Quarto: Translated Homelessness_metrics_county_subgroup.do and Homelessness_metrics_city_subgroup.do into a single parameterized `homelessness.qmd` that handles both county and place geographies
2. Combined historical data pulls: Merged the previously separate 2014-2018 (EdFacts) and 2019-2022 (EdDataExpress) workflows into one document, making future updates simpler
3. Added runner script: `run_homelessness.R` renders both county and place outputs 
4. Fixed testing suite bug: `The evaluate_final_data.R` function was failing when metrics had partial confidence interval coverage (e.g., `count_homeless` has CIs but `share_homeless` does not). Updated the function to check CI alignment per-variable rather than across all columns. This was previously noted in the old `Homelessness Final Data Evaluation.qmd` file.

## Data Notes

- 2023-24 data not yet available on EdDataExpress as of February 2025
- Race/ethnicity breakdowns only available for 2019+; earlier years have total counts only
- Output files:
    - Overall: `homelessness_2014-2022_{county,city}.csv`
    - Subgroups: `homelessness_2019-2022_subgroups_{county,city}.csv`


12_housing-stability/homelessness.qmd
```diff
@@ -0,0 +1,1052 @@
...
+      city_name = city_location |>
+        str_to_lower() |>
+        str_to_title()
+    )
+}
+```
+
+```{r}
+#| label: download-ccd-race
+
```
jwalsh28, line 245: I think this could use a bit more documentation describing the filters in particular. Why do we filter grade and sex for 99? I assume this is a total indicator but would be nice to have that written down.

12_housing-stability/homelessness.qmd
```diff
@@ -0,0 +1,1052 @@
...
+- The "Reset Filters" button does not work properly
+- Filters may disappear or the page may appear blank - refresh and try again
+- Datasets are limited to 150,000 rows
+- The default changed to show only one state (Alabama)
+
+**Note:** As of February 2026, 2023-24 data is not yet available on EdDataExpress.
+
+**Step-by-step instructions:**
+
+1. Go to: https://eddataexpress.ed.gov/download/data-builder/data-download-tool
```
jwalsh28: I found this website to be borderline unusable..
Do we have an expected data for release of the 2023-2024 data?

12_housing-stability/homelessness.qmd
```diff
@@ -0,0 +1,1052 @@
...
+## Clean and Reshape Homelessness Data
+
+```{r}
+#| label: clean-eddataexpress
+
+# Filter to homeless students enrolled (data_group 655) and relevant subgroups
+eddataexpress_clean <- eddataexpress_raw |>
+  filter(
+    data_group == 655,
+    !subgroup %in% c(
```
jwalsh28: Is there a chance the subgroups will change? If so we may want to filter in the explicit subgroups of interest instead of filtering out the ones we do not want.

12_housing-stability/homelessness.qmd
```diff
@@ -0,0 +1,1052 @@
...
+  arrange(desc(year), state, .data[[settings$geo_var]])
+```
+
+## Quality Checks
+
+```{r}
+#| label: quality-checks
+
+# Summary by year
+message("Summary by year:")
```
jwalsh28, line 863: Depending on how difficult this is, it would be nice to see some distributions by subgroup. Maybe even just select a year and show each subgroup to see if anything wacky is happening.

12_housing-stability/homelessness.qmd
```diff
@@ -0,0 +1,1052 @@
...
+  scale_x_continuous(labels = scales::percent_format()) +
+  labs(
+    title = "Distribution of Homeless Share by Year and Quality Flag",
+    x = "Share of Students Experiencing Homelessness",
+    y = "Count"
+  ) +
+  theme_minimal()
+```
+
+## Export Final Data
```
jwalsh28, line 971: Final data is in descending format by year, for data combination we tend to try to keep everything in ascending order. We should arrange by year.

jwalsh28 (commented):
This is a very nice document. Overall I think it could use it a bit more documentation, particularly around data filters and some of the suppression. 

I also added one request on final data format and a suggestion for validation.

Overall looks great!

12_housing-stability/homelessness.qmd
malcalakovalski: Completely agree, the site has serious issues. I visited it again today and the filter sidebar doesn't even render: you get "Your query returned no results" with no way to set filters. 

Claudia documented the same bugs on Box in December 2024 when downloading the 2022-23 data (broken reset button, 150K row limit, disappearing filters, confusing state default).  I've documented the manual download steps in detail based on the doc Claudia wrote and noted that the 2022-23 file is available on Box. 

I haven't found anything on a release date for 2023-2024 data on the dashboard page. I also checked the NCHE page and it shows 2022-2023 as the latest year.

12_housing-stability/homelessness.qmd
malcalakovalski: Good call, I'll add a comment explaining these filter values. In the CCD, `grade = 99` represents the total across all grades and `sex = 99` represents the total across both sexes. Without these filters, the API returns separate rows for each grade and sex combination, which would then need to be aggregated.

12_housing-stability/homelessness.qmd
malcalakovalski: Good catch, I fixed this in my latest round of commits

12_housing-stability/homelessness.qmd
malcalakovalski: Agreed, I switched to a positive filter that explicitly selects the 8 expected subgroups. This is more defensive against future category changes. For reference, the NCHE federal data summary (SY 2019-20 to 2021-22, Table 6) confirms these race/ethnicity categories have been stable across all reported years, but the positive filter is still preferable as a defensive coding practice

12_housing-stability/homelessness.qmd
malcalakovalski: Added a faceted histogram in the visual-checks chunk (line 916) showing the distribution of homeless share by race/ethnicity subgroup

jwalsh28 (approved):
All my initial comments were addressed, this looks good. After rendering should be good to merge.

PR #569, opened 2026-02-13
Update income metric for 2024 ACS and transition files to new folder

jwalsh28:
This metric updates the income metric to produce 2024 results. It also:

1) Moves the metric to a new subfolder "03-opportunities-for-income"
2) Updates the documentation


malcalakovalski (changes requested):
@jwalsh28 This looks good but I've noted a few things that need to be resolved before merging. Otherwise the documentation improvements seem great!

## Critical Issues

1. Broken Path References in Subgroup Files

The subgroup QMD files still reference the **old** `01_financial-well-being` folder instead of the new `03_opportunities-for-income` folder. This causes runtime errors.  This affects `income_metrics_county_sub.qmd`, `income_metrics_place_sub.qmd`, and `run_income_years.qmd`. 

2. Function signature mismatch

The `finalize_metric()` function in `03_opportunities-for-income/R/finalize_metric.R` takes a `.geo_level` parameter but the subgroup files call it without that parameter, causing this script to fail at runtime. 

## Minor Issues

1. Hardcoded year in validation plot: `income_metrics_county_all.qmd` hardcodes `year == 2022` but should use `params$year` for consistency with the place version which uses params$year.

2. Parametrizing the qmd files would improve the maintanability of this metric. 

The four QMD files share about 80-90% of the same code. This would reduce 4 files to 1-2 files and eliminate code duplication, making future updates easier. For example, I noticed the improved documentation in the overall metric quarto files are not present in the subgroup quarto files. Parametrization would reduce the burden of ensuring improvements are made across all four files. I think a pattern like this would be a great improvement:

```
  params:
    year: "2024"
    geography: "county"  # or "place"
    subgroups: FALSE     # or TRUE
```

jwalsh28:
@malcalakovalski A note that I added inflation adjustments to this latest commit which I would appreciate a close review of. If it looks okay I will re-run the subgroup files to add in the same method.

03_opportunities-for-income/income_metrics_county_all.qmd
```diff
@@ -0,0 +1,1051 @@
...
+2015	2015 dollars	X	0.703	=	1999 dollars
+2016	2016 dollars	X	0.694	=	1999 dollars
+2017	2017 dollars	X	0.679	=	1999 dollars
+2018	2018 dollars	X	0.663	=	1999 dollars
+2019	2019 dollars	X	0.652	=	1999 dollars
+2020	2020 dollars	X	0.644	=	1999 dollars
+2021	2021 dollars	X	0.615	=	1999 dollars
+2022	2022 dollars	X	0.569	=	1999 dollars
+2023	2023 dollars	X	0.547	=	1999 dollars
+2024	2024 dollars	X	0.531	=	1999 dollars"
```
malcalakovalski: I'd suggest scraping this table directly instead of copying and pasting for robustness and maintainability. Otherwise, metric leads will have to remember to copy and paste this table across all the scripts. The table is static HTML so `rvest` works cleanly. Here's what the replacement code would look like:


```suggestion
cpi_data <- read_html("https://usa.ipums.org/usa/cpi99.shtml") %>%
    html_element("table") %>%
    html_table() %>%
    janitor::clean_names() %>%
    select(data_year, income_year, cpi_to_1999 = cpi99) %>%
    mutate(
      data_year = as.numeric(data_year),
      income_year = parse_number(income_year),
      cpi_to_1999 = as.numeric(cpi_to_1999)
    ) %>%
    filter(!is.na(data_year)) # drop the second 2000 row which gets parsed as NA due to a trailing space

```

Since this would have to be pasted across files, we could also put it in a standalone R script and write to a csv file. It's up to you! But I think this highlights why the parametrized reports would be useful in the future!

03_opportunities-for-income/income_metrics_county_sub.qmd
```diff
@@ -627,7 +628,8 @@ results_race_ethnicity <- acs_race_ethnicity %>%
 
 ```{r}
 #| label: finalize-race-ethnicity
-results_race_ethnicity <- finalize_metric(results_race_ethnicity)
+results_race_ethnicity <- finalize_metric(results_race_ethnicity,
+                                          .geo_level = place)
```
malcalakovalski:
```suggestion
                                          .geo_level = county)
```

03_opportunities-for-income/income_metrics_county_sub.qmd
```diff
@@ -706,7 +708,8 @@ results_gender <- acs_gender %>%
 
 ```{r}
 #| label: finalize-gender
-results_gender <- finalize_metric(results_gender)
+results_gender <- finalize_metric(results_gender,
+                                  .geo_level = place)
```
malcalakovalski:
```suggestion
                                  .geo_level = county)
```

03_opportunities-for-income/income_metrics_county_sub.qmd
```diff
@@ -550,7 +550,8 @@ Clean up the confidence interval bounds and sort the data.
 
 ```{r}
 #| label: finalize-all
-results_all <- finalize_metric(results_all)
+results_all <- finalize_metric(results_all,
+                               .geo_level = place)
```
malcalakovalski:
```suggestion
                               .geo_level = county)
```

03_opportunities-for-income/income_metrics_place_all.qmd
```diff
@@ -0,0 +1,1050 @@
...
+  filter(st_place %in% c("1714000", "0644000", "4835000", "5167000")) %>% 
+  mutate(place = case_when(place == "14000" ~ "Chicago, IL",
+                            place == "44000" ~ "Los Angeles, CA",
+                            place == "35000" ~ "Houston, TX",
+                            place == "67000" ~ "Richmond, VA"),
+         place = factor(place, levels = c("Chicago, IL",
+                                            "Los Angeles, CA",
+                                            "Houston, TX",
+                                            "Richmond, VA"))) %>% 
+  ggplot(aes(year, pctl_income_50, color = place)) +
```
malcalakovalski: Small bug here: 

```suggestion
  ggplot(aes(year, pctl_income_80, color = place)) +
```

03_opportunities-for-income/income_metrics_county_all.qmd
```diff
@@ -0,0 +1,1051 @@
...
+    toc_float: true
+    embed-resources: true
+    code-fold: show
+execute: 
+  warning: false
+editor_options: 
+  chunk_output_type: console
+params:
+  year: "2024"
+  inflation_adjust: 1.883
```
malcalakovalski: I would document how this parameter is derived. I can see that it's the reciprocal of the 2024 CPI99 factor but it would be good for a future metric lead to know how to update this when new inflation data gets added. 

We could also derive `inflation_adjust` programmatically from the CPI table instead of hardcoding it:

```
inflation_adjust <- 1 / filter(cpi_data, data_year == max(data_year))$cpi_to_1999
```

If we do this, we'd also want to add a `stopifnot()` that the max year matches our expected target year so it doesn't silently use a stale table.

malcalakovalski (changes requested):
I found two minor bugs that should be fixed before merging and have a strong recommendation on how the CPI tables and inflation_adjust are derived. I added suggested code for all of these so it should be quick to update. 

I looked closely and the inflation adjustment methodology and overall the approach is sound!

malcalakovalski (approved):
Looks good to me!

PR #570, opened 2026-02-18
Add simple comparison of local data and data catalog data.

awunderground:


10_construct-database/00_compare-with-data-catalog.qmd
```diff
@@ -0,0 +1,121 @@
...
+
+This sections repeats the comparison for the medians file.
+
+```{r}
+
+medians <- 
+  list(
+    `local` = read_csv(here("10_construct-database", "county_medians_year_suppressed.csv"), show_col_types = FALSE),
+    `catalog` = read_csv("https://urban-data-catalog.s3.amazonaws.com/drupal-root-live/2025/12/12/county_medians_year_suppressed.csv"), show_col_types = FALSE
+  )
```
malcalakovalski: Small syntax error that silently produces wrong behavior: 

-  The `)` closes `read_csv` before `show_col_types`. So `show_col_types = FALSE` becomes a third element of the list() rather than an argument to `read_csv()`.

```suggestion
    `catalog` = read_csv("https://urban-data-catalog.s3.amazonaws.com/drupal-root-live/2025/12/12/county_medians_year_suppressed.csv", show_col_types = FALSE)
  )
```

functions/construction/compare_data.R
```diff
@@ -0,0 +1,155 @@
...
+    vars_diff_catalog <- only_in_catalog |>
+      select(any_of(vars_shared)) |>
+      summarise(across(everything(), ~sum(!is.na(.)))) |>
+      pivot_longer(everything(), names_to = "variable", values_to = "n_rows") |>
+      filter(n_rows > 0) |>
+      pull(variable)
+  } else {
+    vars_diff_catalog <- character(0)
+  }
+
```
malcalakovalski, line 115: I don't think the logic to identify which variables have changed values doesn't do what the output headers claim. It takes the `anti_join` result (rows that differ on any column) and then reports every column that has at least one non-NA value in those rows. This will report nearly every shared column as "different" even when only one column actually changed. 

For example: if a row differs only in `rate_violent_crime`, the anti_join returns the full row, and then every non-NA column (year, state, county, `rate_violent_crime`, etc.) gets flagged as "changed".  This is still helpful for flagging added/removed rows for say, a new year of data but to actually identify which columns differ, we'd need to join on key columns and compare column-by-column:

```
shared <- inner_join(local, catalog, by = c("year", "state", "county"), suffix = c("_local", "_catalog"))
# Then compare _local vs _catalog columns
```

If implementing that is out of scope for now, I'd recommend at minimum renaming the output header to something like "Variables with non-NA values in differing rows" so it doesn't imply column-level diff detection.

functions/construction/compare_data.R
```diff
@@ -0,0 +1,155 @@
...
+#' \dontrun{
+#' data_list <- list(
+#'   local = read_csv("local_file.csv"),
+#'   catalog = read_csv("https://example.com/catalog_file.csv")
+#' )
+#' compare_data(data_list, "mobility-metrics_county.csv")
+#' }
+#'
+#' @export
+compare_data <- function(list, header) {
```
malcalakovalski: The parameter list shadows the base::list() function. Not a bug here, but it makes the code slightly confusing. I'd consider going with something like `data_list` instead

malcalakovalski (changes requested):
Overall, I think this PR makes some great QA improvements with solid documentation but needs some small tweaks:

1. Fix the misplaced parenthesis (bug)
2. Fix or relabel the "variables that differ" section which has misleading output. In particular, the current logic reports all columns with non-NA values in differing rows, rather than the specific columns that changed. Either relabel the output or compare column-by-column on key-matched rows.

Otherwise, my only other suggestion is to consider using a tolerance threshold for comparisons on numeric columns to avoid flagging values that are substantively identical but have some floating-point precision differences due to CSV read/write rounding. This might be a non-issue so we can wait to incorporate this.

functions/construction/compare_data.R
awunderground: Updated

10_construct-database/00_compare-with-data-catalog.qmd
awunderground: Updated

functions/construction/compare_data.R
awunderground: Good catch. I don't know how I let things get so sloppy. 

I updated the code to compare each variable. The issue is that if a new row is added every variable shows up as different (maybe this is fine).

functions/construction/compare_data.R
```diff
@@ -0,0 +1,126 @@
...
+  # Get all variables that exist in both datasets
+  vars_shared <- intersect(names(local), names(catalog))
+  
+  vars_differ <- vector(mode = "logical", length = length(vars_shared))
+  names(vars_differ) <- vars_shared
+  for (var in vars_shared) {
+    
+    vars_differ[var] <- identical(pull(local, var), pull(catalog, var))
+    
+  }
```
malcalakovalski, line 112: I hate to be nitpicky, just wanted to point out this could be a one liner with:

```suggestion
vars_differ <- map_lgl(vars_shared, ~identical(pull(local, .x), pull(catalog, .x)))
```

functions/construction/compare_data.R
malcalakovalski: I think this is desirable and not misleading. Thanks for addressing it!

malcalakovalski (approved):
Looks good to me, thanks for addressing my feedback! I left a comment on replacing a for loop with `map_lgl` but this is mostly a stylistic thing and not needed to merge

functions/construction/compare_data.R
awunderground: I agree. The loop started a little more complex and then ended up simpler. :)

PR #579, opened 2026-03-04
Remove medical debt metric

jwalsh28:
Remove medical debt files and creation in the debt code. 

Remove medical debt files from combination code.


awunderground (approved):
This looks good. I think we should have deleted the final evaluation forms for medical debt.

PR #582, opened 2026-03-25
Update news

awunderground:


NEWS.md
```diff
@@ -1,10 +1,23 @@
+## v2026.01
+
+- Continue migrating metrics into predictor-specific folders.
+- Update share of adults ages 25 to 54 who are employed (#496).
+- Update household income at 20th, 50th, and 80th percentiles (#494).
+- Update number of affordable and available housing units per 100 households with low, very low, and extremely low incomes; and affordable housing units per 100 households w/ low, very low, and extremely low incomes (#505).
+- Update number and share of public-school children who are ever homeless during the school year (#506).
+- Update ratio of the share of total home values owned by a racial or ethnic group to the share of households of the same group (#499).
+- Update share of the voting-age population who turns out to vote (#515).
+- Update average annual improvement in English Language Arts (#501).
```
jwalsh28: this issue is still open, I believe we can move it to complete

jwalsh28: I actually think this is because this metric was not updated, no indication in `tranche2` that any files related with this metric have been changed

jwalsh28 (commented):
Can you just double-check my thoughts on ELA? I might be missing something.

NEWS.md
awunderground: Whoops, this is in tranche3.

jwalsh28 (approved):

PR #591, opened 2026-04-22
Limit precision of combined data

### LINKED ISSUE #545: Limit precision of metrics in the combined database

malcalakovalski:
## Summary

- Adds `round_metrics()` to `combine_functions.R` and calls it before `write_csv()` in all 18 construct-database QMDs
- Rounding happens after all assertions run, so bounds checks use full precision
- Precision by metric type (see [issue comment](https://github.com/UI-Research/mobility-from-poverty/issues/545) for rationale):

| Prefix | Decimals |
|---|---|
| `pctl_income_*` | 0 |
| `share_*` | 4 |
| `rate_*` (except learning) | 1 |
| `rate_learning*` | 2 |
| `ratio_living_wage` | 4 |

- Re-rendered `11_construct_county_all.qmd` and `61_construct_place_all.qmd`. All assertions pass.
- Columns already at target precision (e.g., `rate_injury_death`, `count_*`) are unchanged.

Closes #545


awunderground (changes requested):
I don't think it's right to do the rounding AFTER the assertions because we want the tests to run on the final data. I will move this before the assertions.

awunderground (approved):

PR #592, opened 2026-04-27
Initial commit SEDA

jwalsh28:


07_effective-public-education/MEPS Final Data Evaluation.qmd
awunderground, line 1: Don't include file names with spaces. Ideally, we would move this to the centralized evaluation.

07_effective-public-education/MEPS_metric_city.qmd
```diff
@@ -0,0 +1,407 @@
...
+---
+
+*Program for the creation of the students in high poverty schools metric at the city level*
+
+11/28/2022 
+		Uses MEPS: the share of students attending schools where 20% or more students are in Poverty  **
+		Uses 2014-2018 data
+	  Produces city level data
+10/15/2025 
+		Produces data for 2014-2022 (School Years 2014-15 through 2022-23)
```
awunderground: Let's delete this. We have GitHub for this exact purpose.

07_effective-public-education/MEPS_metric_city.qmd
```diff
@@ -0,0 +1,407 @@
...
+                                      place_name == "Athens-Clarke County unified government (balance)" ~ "Athens",
+                                      place_name == "Augusta-Richmond County consolidated government (balance)" ~ "Augusta",
+                                      place_name == "Macon-Bibb County" ~ "Macon",
+                                      place_name == "Boise City" ~ "Boise",
+                                      place_name == "Lexington-Fayette" ~ "Lexington",
+                                      place_name == "Louisville/Jefferson County metro government (balance)" ~ "Louisville",
+                                      place_name == "Nashville-Davidson metropolitan government (balance)" ~ "Nashville",
+                                      place_name == "McAllen" ~ "Mcallen",
+                                      place_name == "McKinney" ~ "Mckinney",
+         TRUE ~ place_name),
```
awunderground: Why is this needed?

07_effective-public-education/MEPS_metric_city.qmd
```diff
@@ -0,0 +1,407 @@
...
+
+### School level enrollment
+
+Download the CCD school level enrollment by race/ethnicity from Urban's Education Data Portal
+Filter grade for 99 - 99 is the total enrollment in the school, regardless of grade.
+
+Note this download is computationally large, will take time to finish.
+
+```{r}
+input_files <- here::here("08_school-economic-diversity/data/raw/", paste0("ccd_enr_2014-", year_ed, ".rds"))
```
awunderground: I believe this is the wrong file path.

awunderground (changes requested):
Did you delete the old scripts and data?

I found a couple of fatal errors. Back to you.

malcalakovalski (commented):
(see updated review below)

malcalakovalski (changes requested):
It looks like all of Aaron's comments are all addressed: MEPS files removed and old Stata scripts deleted.

The regression translation looks right to me too. Two things though:

1. **Counties/cities without SEDA data are dropped instead of kept as NA rows.** County all file has ~3,000 rows per year instead of 3,143; city all has ~410 instead of 486. This happens because the crosswalk `left_join` produces one row with NA `subgroup` for unmatched counties, and the export filters those out as `subgroup_type == "other"`. To fix this, after the join, you can expand those to all subgroups using `tidyr::crossing()` and then `bind_rows()` back. 

2. **City data includes year 2013.** The cohort filter (`cohort >= 2014`) becomes year 2013 after `year = cohort - 1`. County QMD handles this correctly with a post-adjustment filter. City QMD needs the same.
Also, county QMD doesn't call `evaluate_final_data()` and the county eval forms still say 2014-2018 (2019 is missing). Once those are added, issue #1 would surface immediately.

malcalakovalski (approved):
Both city-level issues from my review are fixed! Once the county file gets updated with the same `expand()` pattern, the `evaluate_final_data()` call, and updates and the eval forms are updated to include 2019, this should be ready to merge!

